PR #4, opened 2018-05-02
[CLI Grader] Add a flag to grade 'student.ml' instead of 'solution.ml'

erikmd:
Dear learn-ocaml developers,

I propose to integrate this feature that allows one to use learn-ocaml in "offline" mode as well.

This feature is disabled by default and just consists in grading `student.ml` instead of `solution.ml`.

For example it can be used with [autograde-ocaml](https://github.com/pfitaxel/autograde-ocaml) (a simple tool we have developed for grading assignments in the "PFITA" teaching unit of Univ. Toulouse 3), which amounts to calling `learnocaml-grader.byte` with the CLI flag `-grade-student`.

### DIFF
--- a/src/grader/grader_cli.ml
+++ b/src/grader/grader_cli.ml
@@ -33,9 +33,14 @@ let display_outcomes = ref false
 (* Where to put the graded exercise *)
 let output_json = ref None
 
+(* Should the tool grade 'student.ml' instead of 'solution.ml' ? *)
+let grade_student = ref false
+
 let args = Arg.align @@
   [ "-output-json", Arg.String (fun s -> output_json := Some s),
     "PATH save the graded exercise in JSON format in the given file" ;
+    "-grade-student", Arg.Set grade_student,
+    " grade file 'student.ml' instead of 'solution.ml'";
     "-display-outcomes", Arg.Set display_outcomes,
     " display the toplevel's outcomes" ;
     "-display-progression", Arg.Set display_callback,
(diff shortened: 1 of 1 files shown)


OCamlPro-Couderc:
This seems reasonable to me, and indeed pretty useful to use learn-ocaml as an offline grader. 

However, it might be useful to use a parameterized argument : `-grade-strudent file.ml`. Using `student.ml` is maybe too restrictive, this means that for each student's code, it must be copied into the same directory as the exercise and renamed `student.ml` (as you do in `autograde-ocaml`), or have as many exercise directories as there is students. 

The `grade_student` reference then contains a `string option`, and is set as `Some file.ml` when the `-grade-student file.ml` option is given, `None` otherwise.

src/grader/grader_cli.ml
```diff
@@ -67,12 +72,22 @@ let remove_trailing_slash s =
   let len = String.length s in
   if len <> 0 && s.[len-1] = '/' then String.sub s 0 (len-1) else s
 
+let read_file f =
```
OCamlPro-Couderc: Since the grader uses Lwt extensively, Lwt_io should be used to read files instead of Pervasives, which should look like : 
```
Lwt_io.with_file ~mode:Lwt_io.Input f Lwt_io.read >>= fun ic ->
Lwt_io.read ic >>= fun content ->
Lwt.return content
```
`with_file` ensures the file descriptor is closed after reading the file.

erikmd:
Hi @OCamlPro-Couderc! thanks for your comments; I've generalized/refactored this PR as suggested.

OCamlPro-Couderc:
Hi @erikmd , thank you for your changes! Merging now.

PR #5, opened 2018-05-02
[test_lib] Fix abstract_type & compatible_type

erikmd:
Hi,
This is just a bugfix PR.
Erik


OCamlPro-Couderc:
Hi Erik,
It seems OK to me, merging now.

PR #6, opened 2018-05-02
[test_lib] Update/Add primitives about types/modules

erikmd:
Dear learn-ocaml developers,

For preparing one of our assignments manipulating modules and abstract types, we have been led to add two functions in `test_lib` (and add an optional argument in function `abstract_type`).

As this could be useful for other users of learn-ocaml, I submit this PR (which relies on #5).

Erik

### DIFF
--- a/src/grader/test_lib.ml
+++ b/src/grader/test_lib.ml
@@ -70,6 +70,9 @@ module type S = sig
 
   val test_student_code : 'a Ty.ty -> ('a -> Learnocaml_report.report) -> Learnocaml_report.report
 
+  val test_module_property :
+    'a Ty.ty -> string -> ('a -> Learnocaml_report.report) -> Learnocaml_report.report
+
   (*----------------------------------------------------------------------------*)
 
   type 'a result =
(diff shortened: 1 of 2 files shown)


OCamlPro-Couderc:
Hi Erik,

Indeed, these functions would have been useful sooner or later. The function `test_module_property` might be enough for the moment, but is too restrictive : since `Introspection` uses the core_types of OCaml, module types are given as first-class module types, which can ony refer to bound module types. This means that one must ensure that there exist a module type in the current environment compatible with the signature of the module we are looking for. As a result, `Introspection.get_value` returns a first-class module. In the future, it might prove useful to have test functions dedicated to modules and signatures, which will deprecate this function.

By the way, I think that adding a `?score` argument for each function that potentially returns a success might be useful in future (not in this PR, of course). Maybe it would have been better to add them in a separate PR, but we can keep the two you added in this one of course.

Sorry for the long post. I think this is mergeable, once the single conflict has been solved ;) (due to the merge of #5)

erikmd:
Hi Pierrick,

Thanks for your feedback! The features of this PR were very important to carry out an assignment with modules and ADTs, but I agree with your comments: the module signature test and the support of optional `score` arguments could be generalized later on.

I've just rebased the PR on latest master and force-pushed.

PR #7, opened 2018-05-02
[Grader] Add individual timeout feature

erikmd:
Hi,

For grading assignments offline easily (without commenting-out code by-hand if some test enters in an infinite recursion…), it was important for us to have a feature of "individual timeout" (i.e., if a test case exceeds a specified amount of time, the grader still terminates but marks the test case as failed).

After some research on ways to achieve this in OCaml, I implemented the solution gathered in this PR (which relies on #4):
- it relies on Unix signals,
- the relevant part of the code is [here](https://github.com/pfitaxel/learn-ocaml/blob/0aa73c318fb94a9003622793ddc9a386491157e5/src/grader/test_lib.ml#L822-L839),
- and it was inspired by [this section of the online book Developing applications with Objective Caml](http://caml.inria.fr/pub/docs/oreilly-book/html/book-ora168.html).

The current limitation of this PR is that it won't necessarily work with js_of_ocaml... (I only tested it in batch mode with [autograde-ocaml](https://github.com/pfitaxel/autograde-ocaml), by using a `-timeout` option that is disabled by default.)

So I guess one could either refine this PR to make it work with js_of_ocaml as well, or merge this PR as-is and add some js_of_ocaml support of individual timeout in a separated PR?
But I'm afraid I couldn't do this extension myself since I'm not js_of_ocaml-savvy enough for the moment.

Kind regards,
Erik

### DIFF
--- a/src/grader/grader_cli.ml
+++ b/src/grader/grader_cli.ml
@@ -36,11 +36,16 @@ let output_json = ref None
 (* Should the tool grade a student file instead of 'solution.ml' ? *)
 let grade_student = ref None
 
+(* Should each test be run with a specific timeout (in secs) ? *)
+let individual_timeout = ref None
+
 let args = Arg.align @@
   [ "-output-json", Arg.String (fun s -> output_json := Some s),
     "PATH save the graded exercise in JSON format in the given file" ;
     "-grade-student", Arg.String (fun s -> grade_student := Some s),
     "PATH grade the given student file instead of 'solution.ml'";
+    "-timeout", Arg.Int (fun i -> individual_timeout := Some i),
+    "INT run each test with the specified timeout (in secs)" ;
     "-display-outcomes", Arg.Set display_outcomes,
     " display the toplevel's outcomes" ;
     "-display-progression", Arg.Set display_callback,
@@ -95,8 +100,9 @@ let grade exercise_dir output_json =
          | None -> Lwt.return (Learnocaml_exercise.(get solution) exo) in
        let callback =
          if !display_callback then Some (Printf.printf "[ %s ]\n%!") else None in
+       let timeout = !individual_timeout in
        code_to_grade >>= fun code ->
-       Grading_cli.get_grade ?callback exo code
+       Grading_cli.get_grade ?callback ?timeout exo code
        >>= fun (result, stdout_contents, stderr_contents, outcomes) ->
        match result with
        | Error exn ->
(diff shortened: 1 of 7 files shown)


erikmd:
(Just rebased this PR as well to fix the conflict with `src/grader/grader_cli.ml`)

AltGr:
Hm, this is not easy to implement reliably: as you mentionned, signals will only work on the offline version, but another concern is that OCaml only catches signals at allocation points: if the student were to write a loop that doesn't allocate (e.g. `let rec f () = f ()`), the alarm signal would actually not be able to interrupt it. See e.g. https://ocaml.github.io/ocamlunix/signals.html#sec86 for details.

I'll check if `lwt` can help here in any way, but it's not likely to make things simpler (esp. since it's purely cooperative by default).

erikmd:
Hi @AltGr
> another concern is that OCaml only catches signals at allocation points: if the student were to write a loop that doesn't allocate (e.g. let rec f () = f ()), the alarm signal would actually not be able to interrupt it.

I don't know for the online/jsoo side, but AFAICT this PR does work to interrupt functions that do no allocation in offline mode, I've just tested it on a similar example… Indeed according to [this thread post referring to OCaml's doc](https://www.reddit.com/r/ocaml/comments/3qapbv/question_about_writing_a_timeout_function_and_the/cwfq1oj/), signals should work in this no-allocation case with the byte-code flavor of OCaml (not the native-code one). So this is a good news because the learn-ocaml evaluator is built upon the byte-code interpreter :-)

But regarding the online version, I just didn't check if `Unix.alarm` is supported in js_of_ocaml. I guess not. And as you suggest, `Lwt` won't probably bring a solution for that…

To sum up, it would be great if this feature can be implemented in both offline and online modes, but IMHO even if this feature can only be supported in the offline version (which is very important due to the impossibility to finely grade manually a student's function that loops on some test-cases, as one can just comment-out the function, which is somewhat unsatisfying…) but not in the online mode, it is not an issue from the student's perspective, as he/she still has the opportunity to change the code and trigger the grading again.

AltGr:
(note: actually, timeout is _already_ handled in the JS workers, so has to give a good return to students who write an infinite loop)

erikmd:
@AltGr 
Thanks!

> (note: actually, timeout is already handled in the JS workers, so has to give a good return to students who write an infinite loop)

You're right: there was already a similar feature on the jsoo side... which is indeed sufficient for the students to detect and avoid writing infinite loops :)
(The main difference w.r.t. this PR was just that the jsoo feature provides a "global timeout" while the CLI feature provides "individual timeouts at test-case level")

PR #8, opened 2018-05-04
Add a Docker image for the minimal server

OCamlPro-Couderc:
Adds the possibilty to build a Docker image containing the minimal server. 

Two scripts are provided:
-scripts/build-docker-image.sh to generate the Docker image
-scripts/scripts/docker-server.sh to run, stop and manage the container


scripts/docker-server.sh
```diff
@@ -0,0 +1,83 @@
...
+           --name "$2" "$3"
+    
+}
+
+while [[ $# -gt 0 ]]; do
+  curr="$1"
+  case $curr in
+      -port)
+      port="$2"
+      shift
```
yurug: I think that a sequence of ``shift`` can be replaced by ``shift 2``.

OCamlPro-Couderc: Indeed, that's fixed now.

README.md
```diff
@@ -58,11 +58,17 @@ you need a local web server.
...
-  You can launch it via
+You can launch it via
 
-```
-./_obuild/*/learnocaml-simple-server.byte
-```
+```./learnocaml-simple-server.byte```.
+
+An Docker container can be generated using this minimal server, via the provided
+script: ```sudo sh scripts scripts/build-docker-image.sh -repo-dir ../learn-ocaml-repository```.
```
yurug: Is it really necessary to be root to build the image? I believed that being in the `docker` group was enough.

README.md
```diff
@@ -58,11 +58,17 @@ you need a local web server.
...
+This generates an image named ```learnocaml-docker``` (by default, but can be
+renamed using the option ```-image-name <my_image>```).
+The server can then be launch using a wrapper around docker CLI:
+```sudo sh scripts/docker-server.sh init```, and the argument ```stop``` to kill
```
yurug: The `init` command fails because you made a typo in the name of image written in `docker-server.sh`:
```
docker_image="learn-ocaml-docker"
```
should be
```
docker_image="learnocaml-docker"
```

OCamlPro-Couderc: My bad, I had a container named that way so I didn't notice the typo. Thanks.

README.md
erikmd: @yurug @OCamlPro-Couderc indeed being just in the `docker` group would suffice, but this could be quite insecure, see e.g. https://fosterelli.co/privilege-escalation-via-docker.html and https://docs.docker.com/engine/security/security/#docker-daemon-attack-surface ; so requiring to be root (using `sudo` or so) is certainly worth it.

Dockerfile
```diff
@@ -0,0 +1,19 @@
...
+
+RUN opam config exec -- opam update
+
+RUN opam config exec -- make build-deps
+
+RUN opam config exec -- make REPO_DIR=exercises_repository
+
+EXPOSE 9090
+
+CMD /home/opam/learn-ocaml/learnocaml-simple-server.byte -port 9090
```
erikmd: According to https://docs.docker.com/engine/reference/builder/#understand-how-cmd-and-entrypoint-interact shouldn't this line be replaced with this?
```
ENTRYPOINT ["/home/opam/learn-ocaml/learnocaml-simple-server.byte", "-port", "9090"]
CMD []
```

Dockerfile
```diff
@@ -0,0 +1,19 @@
+FROM ocaml/opam
+
+WORKDIR learn-ocaml
+
+ADD . .
```
erikmd: It is really a minor detail, but for this line both `ADD . .` and `COPY . .` would do the job, and [this chapter of the doc](https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#add-or-copy) advises to use `COPY` in this case.

erikmd: Also note that doing `COPY . .` will not only copy in the image all source files of the learn-ocaml repo, but also the .git folder. Is it wanted? If you want to avoid this and thereby reduce a bit the size of the image, you could define a `.dockerignore` with a content similar to `.gitignore`, but also with a line `.git`...

Dockerfile
erikmd: Or if the `opam config exec --` prefix is needed and you'd like to rely on [this ENTRYPOINT](https://github.com/ocaml/opam-dockerfiles/blob/5fd5de2854fd944757d3d84ef1fce2604f71631c/Dockerfile#L26), maybe the best choice would be
```
CMD ["/home/opam/learn-ocaml/learnocaml-simple-server.byte", "-port", "9090"]
```

README.md
OCamlPro-Couderc: @erikmd That's right, I added a link to the documentation in the README.

Dockerfile
OCamlPro-Couderc: (I had a `.dockerignore` locally that I forgot to commit, thanks!)

Actually, you remark on `ADD` and `COPY` was pretty useful: the explanation given in the documentation allowed a better caching of the intermediate steps of building the image. In particular, updating the opam repository and installing the dependencies where taking most of the time of the building process (which was quite annoying during the debug). Thank you!

I'll commit it with other modifications soon.

Dockerfile
```diff
@@ -0,0 +1,25 @@
+FROM ocaml/opam
+
+RUN cd /home/opam/opam-repository && git pull
+
+RUN opam config exec -- opam update
+
+WORKDIR learn-ocaml
+
+RUN sudo chown -R opam:nogroup .
```
erikmd: @OCamlPro-Couderc just taking a look at the latest version of the Dockerfile, I was wondering whether it'd be better to move this command 3 lines below instead? (i.e., `RUN sudo chown -R opam:nogroup .` after `COPY Makefile Makefile`?)

scripts/build-docker-image.sh
```diff
@@ -0,0 +1,70 @@
...
+repo_dir="$root"/demo-repository
+image_name="learnocaml-docker"
+
+function print_usage() {
+    printf "Usage: %s <OPTIONS>\n\
+Options:\n\
+  -repo-dir <directory> (default = demo-repository): \n\
+\t Repository containing the exercises.\n\
+  -root <directory> (default = %s): \n\
+\t Root of the learn-ocaml repository.\n" "$0" "$pwd"
```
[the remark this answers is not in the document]
OCamlPro-Couderc: Actually, if `print_usage` is called after parsing at least `-root dir`, the printed default directory will be `dir`, while it should be `$pwd`. However, I'm not sure this option is actually useful and was used as a workaround at the beginning of the development of the docker container. `-repo-dir` can take an absolute path, which is plenty enough.

scripts/build-docker-image.sh
```diff
@@ -0,0 +1,70 @@
...
+          ;;
+      *)    # unknown option
+          option_error $1
+          ;;
+  esac
+done
+
+exercises_repository="$root"/exercises_repository
+      
+if [ -d "$execises_repository" ]; then
```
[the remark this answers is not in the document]
OCamlPro-Couderc: According to my remark on `print_usage`, `"$root"/` should be removed. Right now, it doesn't even accept an absolute path.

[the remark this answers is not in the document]
OCamlPro-Couderc: I deleted my previous comment, since it was related to my previous one on the `-root` option.

I don't think there is a typo here. Before building the docker image, we must copy the exercises into a local temporary directory, which is called `exercises_repository`. I suppose `shellcheck` thinks that line 59, the second `exercises_repository` is a reference to the variable of the same name.

Dockerfile
OCamlPro-Couderc: You're right, actually it is not even useful at that time of the build, and it is done once again once all the files have been into the container. I removed this `chown` in my next commits.

scripts/build-docker-image.sh
```diff
@@ -0,0 +1,70 @@
+#! /bin/sh
```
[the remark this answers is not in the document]
OCamlPro-Couderc: Well spotted, that's fixed.

scripts/build-docker-image.sh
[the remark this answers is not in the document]
erikmd: Thanks for your reply! actually I believe this variable should not be `"$pwd"` nor `"$root"`, but `"$PWD"`? `$pwd` is undefined in my shell.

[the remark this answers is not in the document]
OCamlPro-Couderc: Indeed!

scripts/build-docker-image.sh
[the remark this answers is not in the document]
erikmd: `shellcheck` just said the variable is undefined as `execises` should be `exercises` :)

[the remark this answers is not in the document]
OCamlPro-Couderc: Oh, I should've read more carrefully! Thanks. (I'm too much used to OCaml's static typing)

OCamlPro-Couderc:
Thank you @erikmd and @yurug for the careful reviews! I'm not planning on adding more features or options to the current scripts. I let @AltGr merge it, or I'll do it later if there is no concerns or non-spotted bugs until then.

yurug:
Thanks! I have pushed two commits to fix minor issues but this works great!

PR #13, opened 2018-05-25
Provide a single CLI front-end gathering all the commands

AltGr:
including grading, processing and simple_server.

The existing binaries are not changed at the moment (although the goal
is to remove them at some point), but the new front-end, just
installed as `learn-ocaml`, uses Cmdliner, more consistent options and
inline documentation (see `learn-ocaml --help`). A script for (bash)
shell-completion is even provided: source scripts/complete.sh.

This PR also includes:

- some missing mlis
- a fix for when users claim tokens that the server doesn't know about
- a link to get back to the front page on the exercises page
- an opam2 package definition, which makes it much easier to get all
  the dependencies in place (just run `opam install .`, or `opam
  install . --deps && make build opaminstall` if you prefer to build by
  hand)
- updated some dependencies (the pins are no longer needed)


yurug:
Thanks! This seems very promising! I will review this PR and @OCamlPro-Couderc's PR this week-end.

AltGr:
For review convenience, here is the generated manpage (accessible through `learn-ocaml --help`; the original has some additional colors & formatting)
```
NAME
       learn-ocaml - Learn-ocaml web-app manager

SYNOPSIS
       learn-ocaml [OPTION]... [COMMAND]...

DESCRIPTION
       This program performs various tasks related to generating, serving and
       administrating a learn-ocaml web-app.

COMMANDS
       The COMMAND argument may be one or more of the following. If no
       command is specified, 'build serve' is assumed.

       grade
           Runs the automatic grader on exercise solutions.

       build
           Generates the application based on a repository containing the
           lessons, tutorials and exercises (see REPOSITORY FORMAT).

       serve
           Run a web-server providing access to the learn-ocaml app, as well
           as user file synchronisation.
OPTIONS
       --help[=FMT] (default=auto)
           Show this help in format FMT. The value FMT must be one of `auto',
           `pager', `groff' or `plain'. With `auto', the format is `pager` or
           `plain' whenever the TERM env var is `dumb' or undefined.

       -o DIR, --app-dir=DIR (absent=./www)
           Directory where the app should be generated for the build command,
           and from where it is served by the serve command.

       --repo=DIR (absent=.)
           The path to the repository containing the exercises, lessons and
           tutorials.

GRADER OPTIONS
       --display-outcomes
           display the toplevel's outcomes

       --display-progression
           display grading progression messages

       --display-stdouts
           display the toplevel's standard outputs

       --dump-outputs=PREFIX
           save the outputs in files with the given prefix

       --dump-reports=PREFIX
           save the reports in files with the given prefix

       -e DIRS, --exercises=DIRS
           Directories where to find the exercises to be graded
           (comma-separated). Can be repeated.

       --grade-student=FILE
           grade the given student file instead of 'solution.ml'

       --output-json=DIR
           save the graded exercise in JSON format in the given file

BUILDER OPTIONS
       --contents-dir=DIR
       (absent=/home/lg/ocamlpro/learn-ocaml/_opam/share/learn-ocaml/www)
           directory containing the base learn-ocaml app contents

SERVER OPTIONS
       -p PORT, --port=PORT (absent=8080)
           The TCP port on which to run the server.

       --sync-dir=DIR (absent=./sync)
           Directory where to store user sync tokens.

REPOSITORY FORMAT
       The repository specified by --repo is expected to contain
       sub-directories lessons, tutorials and exercises.

AUTHORS
       Learn OCaml is written by OCamlPro. Its main authors are Benjamin
       Canou, Çağdaş Bozman and Grégoire Henry. It is licensed under the
       GNU Affero General Public License version 3: any instance of the app
       must provide its source code to its users.

BUGS
       Bugs should be reported to
       https://github.com/ocaml-sf/learn-ocaml/issues
```

yurug:
I am not sure wether this interface allows for the following use case. 

1. The teacher deploys a set of exercises on a server. As the source code of the graders must be private, he is the only user to be able to do that.

2. The student grades his answers using a local (compiled) grader obtained from the server.

3. The student uploads his answers and his grade to the server for the teacher to get some feedback about student progression.

I was expecting at least two tools, one for the student to grade and to submit his answers, and another one for the teacher/administrator to publish new exercises.

yurug:
With ``opam 2.0.0~rc2``, under a fresh ocaml 4.05.0 switch and after ``opam install . --deps``, I get:
```
File "_none_", line 1:
Error: Files ./_obuild/toploop_jsoo/toploop_jsoo.cma(Toploop_jsoo)
       and /home/yann/.opam/4.05.0/lib/js_of_ocaml-compiler/js_of_ocaml_compiler.cma(Js_of_ocaml_compiler__Util)
       make inconsistent assumptions over interface Js_of_ocaml_compiler__Util
```
My distribution is a debian testing.

AltGr:
This is when running `make` ? Maybe just clean up the build (`rm -rf  _obuild`) ?

AltGr:
> 1. The teacher deploys a set of exercises on a server. As the source code of the graders must be private, he is the only user to be able to do that.
> 2. The student grades his answers using a local (compiled) grader obtained from the server.
> 3. The student uploads his answers and his grade to the server for the teacher to get some feedback about student progression.

I am not sure we perfectly understand each other on 2.: if by compiled, you mean "to javascript", then it's what happens now. If you mean "to native code", and assuming the students are running the same arch/system as the teacher and server, I think that could be done, but that's quite outside the picture I had for the system, so we probably should discuss in more detail!

In other words, in my picture there are indeed two parts, the _program_, which is used for 1., and the _web-app_, which is used for 2. — and could be used for 3. as well, but that's not done yet. It is technically a compiled grader obtained from the server and running locally on the student machine, but it's running within the browser.

yurug:
Yes, when I wrote "a compiled grader", I meant "to javascript". Native compilation has never been considered for now and is not a priority.

Now I see that this tool is meant for teachers. I was mislead by the "grade" command which is here used to grade the solutions of the teacher or the solutions of a student.

I guess we need another tool for students that want to grade their solutions built using their favorite editor (emacs or vim) without forcing them to cut and paste their code in a browser.

AltGr:
Ok. The `grade` command as it is provided in this PR is equivalent to what the `learnocaml-grader.byte` was doing (and is still doing, I didn't remove anything yet, just added the more convenient layer).

For the students wanting to grade from the command-line, I guess it's not as high in the priorities, but we have a few options:
- handle grading server-side, and provide a simple API call. Would be nice and safer, but maybe a little costly on the server
- provide them with a generic grader exec (possibly pre-compiled for different systems), that needs both a grader content and a solution. The "grader content" could be obtained from the server directly, and should clearly not be plain code
- generate a grader binary instance that has the grader content for the given exercises embedded. Probably more easily done as an extension of the point above
- a fancy way for a CLI tool to interact with the app running in the local browser ? Might be possible using WebRTCDataChannels...

yurug:
"Grading from the command-line" is high in the priorities since students of Paris Diderot will preferably use that interface to learn-ocaml. (They are not beginners in programming as they are third year students.)

About possible options to implement it, server-side grading is not really satisfying because I would love to follow a scheme that scales well. 

Would it be technically possible to have a daemon that imports the javascript code of the grader from the server and runs this javascript code in a standalone javascript interpreter like nodejs or spidermonkey?

yurug:
@AltGr Does this technical proposal make sense?

AltGr:
Point 2 above seems like the best approach. Rather than importing javascript code, since it's generated from OCaml anyway, we could provide a client (in bytecode) that downloads the exercise from the server (already available as json), and runs the grading client-side. Just like what the javascript client does, actually, so it's just writing the CLI interface, and porting the code that is specific to jsoo (e.g. `XmlHTTPRequest` and uses of the browser's Json parser).

On it. Mockup:
```
$ learn-ocaml-client --server https://learn-ocaml.org exercise-id.ml
> Fetching https://learn-ocaml.org/exercises/exercise-id.json
> Decrypting & running the specific grader
> Report: XX% complete
<full report>
> Submitting solution & grade back to server
```

PR #15, opened 2018-05-30
This changes the Docker file to require a repository, use opam 2, and use it for running Travis tests

AltGr:
Note: this is based on the ocaml/opam2-staging docker images, which are not stable yet.
So I don't recommend merging right away.


AltGr:
Note also that I have not yet ported Pierrick's scripts for easier server management

PR #16, opened 2018-05-30
Update a little the JS libs

AltGr:
- update to the newer ACE, that has a few bugfixes, and might provide
  features we'll want to use in the future. Ported a required fix that
  we had for the earlier version
  
- fix the URL for Mathjax CDN, which is no longer maintained (we need
  to consider embedding a stripped down version if we want it to work
  without Internet access, too. It's the only thing that's not
  embedded at the moment. See the individual commits for pointers on
  how that could be done)


yurug:
Please open an issue about the embedding of MathJax.

PR #20, opened 2018-06-01
Fix broken paths after #14

AltGr:


AltGr:
My bad, I missed those and didn't detect it because I didn't clean everything. Should be back to working with this patch.

PR #22, opened 2018-06-01
Internationalisation (of the interface)

AltGr:
This adds:
- an internationalisation engine (ocplib_i18n + ppx_ocplib_i18n) ; see
  https://github.com/OCamlPro/ocplib-i18n
- wrappers for strings that get displayed to the user (using the ppx; I
  might have missed some)
- a french translation
- automatic choice of the translation based on the browser setting (with
  a fallback to English)

How it works:

- there is a very small PPX that expands `[%i"test"]` to
  `(Ocplib_i18n.s_ [|"transl_0";...;"transl_n"|])` (`[%if"text"]` is
  also supported for format strings)

- `Ocplib_i18n` just provides:
  - a function to select the active lang
  - the `s_` function above that selects the correct translation from
    the string arrays

- translations are read from a LANG.po file in `translations/`. The
  format is a subset of the `gettext` format, which means there is lots of
  tooling available for it (including a powerful emacs mode)

- warnings get displayed if some strings don't have translations. When
  that happens, run `make update-LANG-translation` to synchronise the
  `translations/LANG.po` file, then open it to add the missing
  translations

- to add a new language, simply run `touch translations/LANG.po && make
  update-LANG-translation`

- for strings that appeared statically in the base `html` files, I added
  `txt_xxx` idents, and the base JS code of the page is in charge of
  replacing the corresponding `innerHtml`.


src/app/learnocaml_index_main.ml
```diff
@@ -637,23 +637,14 @@ let set_string_translations () =
...
-let set_lang () =
-  match Js.Optdef.to_option (Dom_html.window##.navigator##.language) with
-  | Some l -> Ocplib_i18n.set_lang (Js.to_string l)
-  | None ->
-      match Js.Optdef.to_option (Dom_html.window##.navigator##.userLanguage)
-      with
-      | Some l -> Ocplib_i18n.set_lang (Js.to_string l)
-      | None -> ()
-
 let () =
```
yurug: The similarity between `learnocaml_index_main.ml` and `learnocaml_exercise_main.ml` is striking. Can these two files share more code?

Makefile
```diff
@@ -56,11 +56,29 @@ learn-ocaml.install:
...
 
+# Generates up-to-date translation template for lang % from the sources
+LANGS = $(patsubst translations/%.po,%,$(wildcard translations/*.po))
+translations/$(LANGS:=.pot):
+	@for f in $(LANGS); do echo >> translations/$$f.po; done
+	@rm -f translations/*.pot
+	@DUMP_POT=1 ocp-build -j 1
+	@for f in $(LANGS); do \
+	  mv translations/$$f.pot translations/$$f.pot.bak; \
+	  msguniq translations/$$f.pot.bak > translations/$$f.pot; \
```
yurug: Is GNU/gettext standard on MacOS X?

AltGr: I don't know really, but it certainly can be installed without too much effort (e.g. using [homebrew](https://formulae.brew.sh/formula/gettext)). It is, in any case and mot importantly, not needed in the normal build process, but only to synchronise the translations after a change of the strings in the source, or to generate a new translation template to start a new language.

OCamlPro-Couderc:
I think we can merge it, if there is no objection. I agree with @AltGr, `gettext` is available in homebrew and I suppose OSX users are used to it in a development environment.

PR #38, opened 2018-06-21
Small fixes & improvements

AltGr:
Fix the completion script, '--grade-student' option, organise www/

### DIFF
--- a/scripts/complete.sh
+++ b/scripts/complete.sh
@@ -2,7 +2,7 @@ _learn_ocaml_add_f()
 {
   local cmd
   cmd=$1; shift
-  _learn_ocaml_add "$($cmd "$@" 2>/dev/null)"
+  IFS=$'\n' _learn_ocaml_reply+=("$($cmd "$@" 2>/dev/null)")
 }
 
 _learn_ocaml_flags()
--- a/src/grader/grader_cli.ml
+++ b/src/grader/grader_cli.ml
@@ -96,7 +96,7 @@ let grade exercise_dir output_json =
        let exercise_dir = remove_trailing_slash exercise_dir in
        read_exercise exercise_dir >>= fun exo ->
        let code_to_grade = match !grade_student with
-         | Some path -> read_student_file exercise_dir path
+         | Some path -> read_student_file (Sys.getcwd ()) path
          | None -> Lwt.return (Learnocaml_exercise.(get solution) exo) in
        let callback =
          if !display_callback then Some (Printf.printf "[ %s ]\n%!") else None in
--- a/src/repo/learnocaml_process_exercise_repository.ml
+++ b/src/repo/learnocaml_process_exercise_repository.ml
@@ -147,6 +147,10 @@ let main dest_dir =
     match !exercises_index with
     | Some exercises_index -> exercises_index
     | None -> !exercises_dir / "index.json" in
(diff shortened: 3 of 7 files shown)


scripts/complete.sh
```diff
@@ -2,7 +2,7 @@ _learn_ocaml_add_f()
 {
   local cmd
   cmd=$1; shift
-  _learn_ocaml_add "$($cmd "$@" 2>/dev/null)"
+  IFS=$'\n' _learn_ocaml_reply+=("$($cmd "$@" 2>/dev/null)")
```
yurug: Redefinition of IFS is very fragile. (I am working on the semantics of POSIX shell in the COLIS ANR project and I would not recommend playing with IFS.)

src/grader/grader_cli.ml
```diff
@@ -96,7 +96,7 @@ let grade exercise_dir output_json =
        let exercise_dir = remove_trailing_slash exercise_dir in
        read_exercise exercise_dir >>= fun exo ->
        let code_to_grade = match !grade_student with
-         | Some path -> read_student_file exercise_dir path
+         | Some path -> read_student_file (Sys.getcwd ()) path
```
yurug: Is that a bugfix?

src/repo/learnocaml_process_exercise_repository.ml
```diff
@@ -147,6 +147,11 @@ let main dest_dir =
     match !exercises_index with
     | Some exercises_index -> exercises_index
     | None -> !exercises_dir / "index.json" in
+  let exercises_path =
+    dest_dir / Filename.dirname (Learnocaml_index.exercise_path "foo") in
```
yurug: Why "foo"?

src/repo/learnocaml_process_tutorial_repository.ml
```diff
@@ -63,6 +63,11 @@ let main dest_dir =
     match !tutorials_index with
     | Some tutorials_index -> tutorials_index
     | None -> !tutorials_dir / "index.json" in
+  let tutorials_path =
```
yurug: This duplication frightens me.

src/grader/grader_cli.ml
AltGr: yes, previously the argument would be checked as an existing file in CWD, then used as a path relative to `exercise_dir`. This makes the two consistent, with a more standard interface.

src/repo/learnocaml_process_exercise_repository.ml
AltGr: any non-empty string would do, since we extract the dirname. `foo` is I believe a standard place-holder for such cases. I agree this is not very pretty, and we could either check for an existing exercise, or export the dirname directly from `Learnocaml_index`.

src/repo/learnocaml_process_tutorial_repository.ml
AltGr: You mean `tutorials_dir` vs `tutorials_path`? It's poor naming actually... The global reference `!tutorials_dir` is the configuration of the tutorial _source_ directory, while `tutorials_path` points to the destination directory under `www`. I'll change for more explicit names.

AltGr:
Amended my patch for less confusing variable naming, and better inner interface.

PR #40, opened 2018-06-21
docs/: Document meta.json version 2.

yurug:
Here is a proposal for the new version of the format for `meta.json` files.


docs/exercises_format.md
```diff
@@ -60,9 +60,34 @@ An exercise is described by a directory containing at most the following files:
...
+  /* The skills and concepts that are required to do this exercise. */
+  "requirements"       : ["skill1", ..., "skillN", ..., "concept1", ..., "conceptM"],
+  /* The suggested exercises in case of success */
+  "forward_exercises"  : [ "exercise1", "exercise2", ... ],
+  /* The suggested exercises in case of difficulty */
+  "backward_exercises" : [ "exercise1", "exercise2", ... ],
+  /* Comments (in markdown) by the authors or users of this exercise. */
+  "comments"           : ["comment1", "comment2", ...],
+  /* Number of times the exercise has been checked out. */
+  "popularity"         : 1234
```
OCamlPro-Couderc: "comments" and "popularity" cannot be completely static since they are dependent on the usage of the current instance of the platform. However, these fields could be used as the starting value for these entries in a database that is used for statistics or exercises improvements.

These fields could indeed be the result of exporting an exercise from the platform.

docs/exercises_format.md
```diff
@@ -60,9 +60,34 @@ An exercise is described by a directory containing at most the following files:
...
+Json description in version "2" contains more metadata:
+```
+{
+  "learnocaml_version" : "2",
+  "kind"               : "exercise" | "problem" | "project",
+  "stars"              : [1 .. 5],
+  /* In an exercise repository, each exercise must have a unique identifier. */
+  "identifier"         : "some_unique_identifier",
+  /* Authors with their emails. */
+  "authors"            : ["Xavier Leroy <some@email>", "Damien Doligez <someother@email>"],
```
OCamlPro-Couderc: In terms of implementation, maybe the author field could be a tuple (this is simply to avoid parsing the `< .. >` as the email field, which is indeed quite simply).

docs/exercises_format.md
yurug: Yes, you are right. I remove them.

docs/exercises_format.md
yurug: Good idea. I fix that.

yurug:
@OCamlPro-Couderc : if you accept this format, could you start adapting the parser?

PR #44, opened 2018-06-27
Update `meta.json` encoding

OCamlPro-Couderc:
Follows #40. 

However, this encoding is temporary. It allows to use any field in `meta.json` which is obviously not robust. This means that it will not warn for errors if a field is declared as optional and written with a typo.

### DIFF
--- a/src/repo/learnocaml_process_exercise_repository.ml
+++ b/src/repo/learnocaml_process_exercise_repository.ml
@@ -73,7 +73,7 @@ let exercise_meta_enc =
           (opt "requirements" (list string))
           (opt "forward" (list string))
           (opt "backward" (list string)))
-       unit)
+       unit) (* FIXME: temporary parameter, that allows unknown fields *)
 
 let opt_to_list_enc = function
     None -> []
--- a/src/repo/learnocaml_process_exercise_repository.ml
+++ b/src/repo/learnocaml_process_exercise_repository.ml
@@ -42,7 +42,7 @@ let index_enc =
           (fun (title, map) -> (title, `Groups map)) ] in
   let open Json_encoding in
   mu "group" @@ fun group_enc ->
-  check_version_1 @@
+  check_version_2 @@
   union
     [ case
         (obj1 (req "exercises" (list string)))
@@ -203,7 +203,14 @@ let main dest_dir =
                   let exercise =
                     { exercise_kind ; exercise_stars ;
                       exercise_title = Learnocaml_exercise.(get title) exercise ;
-                      exercise_short_description} in
+                      exercise_short_description;
+                      exercise_identifier ;
+                      exercise_author = opt_to_list_enc author ;
+                      exercise_focus = opt_to_list_enc focus ;
+                      exercise_requirements = opt_to_list_enc requirements ;
+                      exercise_forward = opt_to_list_enc forward ;
+                      exercise_backward = opt_to_list_enc backward ;
+                    } in
                   acc >>= fun acc ->
                   Lwt.return (StringMap.add id exercise acc))
                (Lwt.return StringMap.empty) ids >>= fun exercises ->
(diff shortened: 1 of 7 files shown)


src/repo/learnocaml_process_exercise_repository.ml
```diff
@@ -182,14 +194,23 @@ let main dest_dir =
...
-                  >>= fun (exercise_kind, exercise_stars) ->
+                  >>= fun ((exercise_kind, exercise_stars, exercise_identifier,
+                            author, focus, requirements,
+                            forward, backward), _) ->
                   let exercise_short_description = None in
                   let exercise =
                     read_exercise (!exercises_dir / id) in
                   let exercise =
                     { exercise_kind ; exercise_stars ;
                       exercise_title = Learnocaml_exercise.(get title) exercise ;
```
yurug: Do you know why the exercise title is not part of the meta.json file? 

If this is the only place where the title.txt file is read, it could be a good idea to first look into meta.json for a title and if there is no such field, look into title.txt.

AltGr: I am not sure... maybe the `title.txt` is another remnant of the `FUN` platform, but I don't see a reason now to keep it separate indeed.

lyrm: Actually, there are no title.txt files in the FUN exercises (or at least, in the version I have) and the title is in the meta.json file.
An example :
{ "version": 1,
  "kind": "exercise",
  "title": "Displaying a Filesystem Hierarchy", 
  "stars": 3 }.

OCamlPro-Couderc: Indeed, it might be a good idea to remove this file in this PR and merge it with `meta.json`. Another improvement would be the `short_description` field from `Learn_ocaml_index.exercise`, which is always set to `None` for the moment.

OCamlPro-Couderc: Another file that could be easily integrated is `max_score.txt`, I do no see a good reason to keep it separated for the moment.

yurug: @OCamlPro-Couderc Would you mind implement this feature if it is not already there, please? (The title in `meta.json` by default and in `title.txt` as a fallback.)

@lyrm Can you update the tutorial once the modification is done, please?

yurug: @OCamlPro-Couderc Let's integrate `max_score.txt` to `meta.json` too.

src/repo/learnocaml_process_exercise_repository.ml
```diff
@@ -194,15 +209,22 @@ let main dest_dir =
...
                   let exercise =
                     read_exercise (!exercises_dir / id) in
+                  let exercise_title =
+                    match title with
+                    | Some title -> title
+                    | None -> Learnocaml_exercise.(get title) exercise
+                  in
                   let exercise =
                     { exercise_kind ; exercise_stars ;
-                      exercise_title = Learnocaml_exercise.(get title) exercise ;
```
yurug: Is there a fallback to reading `title.txt` if the field is not here?

OCamlPro-Couderc:
`title.txt` and `max_score.txt` are integrated into `meta.json`. If the field is not available, it falls back to the said file (`max_score` is completely optional).

yurug:
Can you update the docs please?

lyrm:
About max_score : I have actually find what I think is a bug with max_score. I am trying to fix it (or will open an issue if I can't) but basically, I have an exercise where max_score does not match the actual maximum possible score for the exercise (3 functions are tested, for 10 points each. max_score = 20 instead of 30). 

And why put it in meta.json ? Can it be actually changed ? Isn't it a computed value, depending of the number of tests done ?

OCamlPro-Couderc:
Actually, `max_score` is never used (it wasn't even in the exercise representation used by the grader, so it couldn't access it). I suppose this was imagined to override the grader's given score if it returns a score higher than `max_score`. As such, it is not computed.

I think it should be easy to implement now that it is available into the representation of exercises.

PR #46, opened 2018-07-03
Correction of behavior of the optional argument for the test_function…

lyrm:
The optional argument ~gen of the test_function had  a pretty bad default behavior : 
- if not defined the number of automatically generated tests was **NOT** 0
- if ~gen:0 :  the grader still checked if there was the right sampler to generate tests automatically which was ... pretty inconvenient.

Now : if ~gen is not defined, the number of generated tests is 0 and if not defined or if ~gen=0 there is no need to define a sampler for the inputs of the tested function.

NB: The change of the default value of ~gen will break the mooc exercises, so maybe we should discuss it!

### DIFF
--- a/src/grader/test_lib.ml
+++ b/src/grader/test_lib.ml
@@ -1196,7 +1196,7 @@ let run_timeout ~time v =
     test_value rf @@ fun rf ->
     let gen = match gen with
       | Some n -> n
-      | None -> 0 in
+      | None -> max 5 (10 - List.length tests) in
     let tests = match gen with
       | 0 -> List.map (fun x () -> x) tests
       | _ -> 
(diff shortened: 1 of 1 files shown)


erikmd:
@lyrm This PR is very welcome for us as well! (as we found ourselves needing to pass some dummy sampler `let dummy : 'a sampler = fun () -> failwith "dummy sampler"` for several tests having `~gen:0` about user-defined types, so IINM we won't need this workaround anymore :)

AltGr:
I guess the policy behind this was that exercises should always have sampled tests ? But in any case the default value should have been documented.

Otherwise, couldn't a student just hard-code the expected output for all the known test-cases ?

erikmd:
To follow-up @AltGr's comment, maybe a trade-off could be considered? namely adapting this PR to:
- keep the default value of `~gen:(max 5 (10 - List.length tests))` and document it,
- but avoid raising `Exception: Failure "unsamplable type".` for `~gen:0` when no `~sampler` argument is passed?

AltGr:
@erikmd: that sounds perfect, yes

Issue #50, opened 2018-07-04
Question: construct a prot term from the string representation of the type


erikmd:
### Context
With the current API of test_lib, fixed-arity primitives such as `test_function_1` take as argument some `_ Ty.ty`:
```
val test_function_1 : ?test: 'b tester -> …
  ('a -> 'b) Ty.ty -> string -> ('a * 'b * string * string) list -> Learnocaml_report.report
```
while the n-arity primitives take as argument some `(_, _, _) prot`, the advantage of which being thus that the co-domain type `'ret` is clearly specified:
```
val test_function : ?test: 'ret tester -> …
  (('ar -> 'row) Ty.ty, 'ar -> 'urow, 'ret) prot -> ('ar -> 'row) lookup ->
  (('ar -> 'row, 'ar -> 'urow, 'ret) args * (unit -> 'ret)) list -> Learnocaml_report.report
```

In learn-ocaml-editor we decided to preferably rely on `test_function` (and `test_function_against`) to have more generic code, while providing notations to easily construct some term of type `(('ar -> 'row, 'ar -> 'urow, 'ret) args * (unit -> 'ret)) list`.

The current bottleneck for specifying such test cases is that the syntax to construct a term of type `(('ar -> 'row) Ty.ty, 'ar -> 'urow, 'ret) prot` is not very user-friendly. We thus obtain some code like this:

```
type test_qst_typed =
  | TestAgainstSol :
      { name: string
      ; prot: (('ar -> 'row) Ty.ty, 'ar -> 'urow, 'ret) prot
      ; gen: int
      ; suite: ('ar -> 'row, 'ar -> 'urow, 'ret) args list } -> test_qst_typed
  …
  | TestSuite :
      { name: string
      ; prot: (('ar -> 'row) Ty.ty, 'ar -> 'urow, 'ret) prot
      ; suite: (('ar -> 'row, 'ar -> 'urow, 'ret) args * (unit -> 'ret)) list } -> test_qst_typed

let example_constr_sol =
  TestAgainstSol
    { name = "min";
      prot = (arg_ty [%ty: int] (last_ty [%ty: int] [%ty: int]));
      gen = 4;
      suite = []
    }
```

### Question
The question is: would it be feasible to add to the test_lib API a facility to get a prot from a string representation of the type? Basically one of the two approaches would be OK for us:

1. Either implement a syntax `[%prot: a -> b]` similar to `[%ty: a -> b]`
[...]

PR #57, opened 2018-07-10
test_lib.mli: Add another way to build ['a Ty.ty] terms

erikmd:
Dear learn-ocaml developers,

To follow-up our latest discussion, I'm opening this PR to propose another way to construct `'a Ty.ty` terms so that the grader fails more gracefully in case there is an undefined type.

This corresponds to the following use case: the type of a function to grade depends on a type `t` that
should be defined in the student's Code itself.

If that type is undefined, `[%ty: Code.t]` will trigger a syntax error (cf. the first screenshot below) while `(ty_of "Code.t")` will fail locally only, and let the other tests pass (cf. the second screenshot below).

Kind regards, Erik

### Using `[%ty: Code.t]`
screenshot_learn-ocaml-editor_ty_ppx

### Using `(ty_of "Code.t")`
screenshot_learn-ocaml-editor_ty_of

### DIFF
--- a/src/grader/test_lib.ml
+++ b/src/grader/test_lib.ml
@@ -257,6 +257,8 @@ module type S = sig
 
   (*----------------------------------------------------------------------------*)
 
+  val ty_of : string -> 'a Ty.ty
+
   (* Usage: arg 3 @@ arg "word" @@ last false *)
   type ('arrow, 'uarrow, 'ret) args
   val last :
@@ -1102,6 +1104,8 @@ let run_timeout ~time v =
     | Last_ty (a, b) -> Ty.curry a b
     | Arg_ty (x, Last_ty (l, r)) -> Ty.curry x (Ty.curry l r)
     | Arg_ty (x, Arg_ty (y, r)) -> Ty.curry x (ty_of_prot (Arg_ty (y, r)))
+
+  let ty_of str = Ty.repr (Parse.core_type (Lexing.from_string str))
 
   let rec apply : type p a c r. (p -> a) -> (p -> a, p -> c, r) args -> r = fun f x ->
     match x with
--- a/src/grader/test_lib.mli
+++ b/src/grader/test_lib.mli
@@ -257,6 +257,13 @@ module type S = sig
 
   (*----------------------------------------------------------------------------*)
 
+  (** Parsing function.
+
+      [ty_of "string"] is very similar to the ppx expr [[%ty: string]]
+      but [ty_of "Code.t"] won't trigger a syntax error if the type t
+      is undefined in student's Code. *)
+  val ty_of : string -> 'a Ty.ty
+
   (** The type of arguments, represented as heterogeneous lists.
 
   Usage: [arg 3 @@ arg "word" @@ last false] *)


AltGr:
Unfortunately, the bigger difference between the two is that the second is entirely resolved at run-time, whithout any typing: we effectively obtain a `string -> 'a Ty.ty` function, which defeats the purpose of having a `Ty.ty` type in the first place, and means that the student function will be applied to the testing samples in a completely untyped way. We don't care that the `Code.t` type was defined since we use `'a` in its place...

Maybe if you could give a more concrete example of what you want to achieve we could try to find a better solution? I have trouble seeing how you can test a function against known or generated values of a type you don't know about. Maybe you just want to check a given module interface and then run the actual tests on a combination of functions (e.g. test `s  -> cardinal (t_of_string s)` rather than `t -> cardinal t` directly) ?

It's actually weird that the first error didn't happen at build time, maybe the bug was actually that it could see the type defined in the solution ?

yurug:
I understand that we should provide a way for teachers to minimize the dependencies between exercises questions. However, as the solution of the student is provided in a single OCaml file, it is mandatory in OCaml to at least provide a file that typechecks entirely, so as @AltGr I do not see what you want to achieve exactly.

erikmd:
Hi @AltGr and @yurug,
Thanks a lot for your comments! and sorry for late reply.

> It's actually weird that the first error didn't happen at build time, maybe the bug was actually that it could see the type defined in the solution ?

Good question: I've just tested that `make process-repo` indeed makes the first error appear at build time.

I had not tested this at first as the screenshot was taken from the learn-ocaml-editor "Grade" button in teacher mode (which processes the grader on the solution) and I thought that `make process-repo` was doing the same...

So IIUC, `make process-repo` only checks that every .ml file compiles, separately, without e.g. evaluating `test.ml` on the provided `solution.ml`. (But the teacher will still be able to do that latter verification by using `./_obuild/learnocaml-grader/learnocaml-grader.byte` or learn-ocaml-editor.)  

> Unfortunately, the bigger difference between the two is that the second is entirely resolved at run-time, whithout any typing: we effectively obtain a `string -> 'a Ty.ty` function, which defeats the purpose of having a `Ty.ty` type in the first place, and means that the student function will be applied to the testing samples in a completely untyped way. We don't care that the Code.t type was defined since we use 'a in its place...

You're right: the function I proposed to add in this PR is actually untyped, so the corresponding errors (if `Code.t` or so is unbound) will only show up at runtime (not when running `make process-repo`). But it was the simplest solution we found to implement a use case involving abstract types (see below for more info). 

> Maybe if you could give a more concrete example of what you want to achieve we could try to find a better solution? I have trouble seeing how you can test a function against known or generated values of a type you don't know about. Maybe you just want to check a given module interface and then run the actual tests on a combination of functions.

Exactly! :)
To be more precise, I have developed some assignments where the student must:
- start with a module signature SIG given by the teacher containing an *abstract type* `t`
- develop a `module MOD : SIG` satisfying this signature
- develop some functions that rely on MOD functions
- develop several other functions that do not rely on MOD

On the teacher side, a requirement was that the `test.ml` grader should compile even if the module `MOD` is not there, in order to provide some feedback for the other functions provided by the student.

So there are two problems here:
1. the type at stake (here, `MOD.t`) is abstract and defined by the student, basically the `test.ml` file cannot construct terms of this abstract type directly. 
2. the `[ty: MOD.t]` ppx invocation is resolved at build time, so it won't address the requirement above in case of absence of module `MOD` (compilation error).

I had solved point 1. by combining several functions asked to the student, exactly as @AltGr mentioned with the `cardinal` example. To sum up it is feasible (but a bit verbose) to implement this with the current test_lib API. Maybe we could discuss this later on to see if the API could be slightly extended to implement this more easily.

So this PR is mostly about point 2.
Do you believe we could find a better solution than relying on this function `ty_of` ?

Kind regards, Erik

AltGr:
Yes, if possible, I would prefer to avoid the `ty_of` solution — sorry!

Being able to test whole modules in `test_lib.ml`, just like we test functions, would be quite nice indeed. I wonder if we couldn't cheat to get that easily with first-class modules: after all testing `Mod` can be achieved by testing `fun (module Mod: SIG) -> xxx`

On a related subject, It may not be too difficult to define a `[%prot ]` ppx extension similar to `[%ty ]`, but returning the `prot` type.

I'll look into these.

erikmd:
Hi @AltGr, thanks for your feedback!

Following your suggestion, I just made a small test based on the code
```
test_module_property [%ty: (module SIG)] "MOD" (fun (module Mod : SIG) -> ...
```
and it appears to fully address the point 1. I had mentioned in my last comment, and would lead to less verbose code w.r.t. my first implementation. Thanks!

And regarding the implementation of point 2. (writing a test for a function that is outside `MOD`, but whose type relies on `MOD.t`, which may be undefined (in which case a ppx like `[%ty: MOD.t -> string list]` wouldn't compile)), I just realized than one can rely on first-class modules as well, and require that the student provides a functor... I have just tested this and it works like a charm! :) So I can close this PR which now becomes useless.

> On a related subject, It may not be too difficult to define a `[%prot ]` ppx extension similar to `[%ty ]`, but returning the prot type.

OK :) the availability of one such feature described in #50 would be awesome for us!

erikmd:
@AltGr thanks for your comment!

> Of course, more effort would be required to run properly generated tests, test the code for exceptions, and generate a more informative report. This is where test_lib might be lacking:

Yes I agree.

But I am not sure it would be practical to implement and use a function such as `test_module_against_solution`.
AFAICT, for testing a module that is an "abstract data type", I would find more natural to test this module against its specification, i.e. by running assertions that would correspond to the "axioms" of the ADT, rather than providing the teacher with two opaque modules (the student's and the solution's module).

So I would propose the following:
Add functions `test_immediate_*` in `test_lib`, similar to `test_function_*` but without the lookup:

```ocaml
val test_immediate_1 :
    ?test: 'b tester ->
    ?test_stdout: io_tester ->
    ?test_stderr: io_tester ->
    ?before : ('a -> unit) ->
    ?after : ('a -> ('b * string * string) -> ('b * string * string) -> Learnocaml_report.report) ->
    ('a -> 'b) ->
    ('a * 'b * string * string) list -> Learnocaml_report.report

val test_immediate :
    ?test: 'ret tester ->
    ?test_stdout: io_tester ->
    ?test_stderr: io_tester ->
    ?before : (('ar -> 'row, 'ar -> 'urow, 'ret) args ->unit) ->
    ?after :
      (('ar -> 'row, 'ar -> 'urow, 'ret) args ->
       ('ret * string * string) ->
       ('ret * string * string) ->
       Learnocaml_report.report) ->
    ('ar -> 'row) ->
    (('ar -> 'row, 'ar -> 'urow, 'ret) args * (unit -> 'ret)) list -> Learnocaml_report.report
```

So we could just use `test_module_property` and locally build "abstract test cases" that are then passed to the `test_immediate` function.

For example, for testing a stack ADT, I would write something like this:

```ocaml
(* prepare.ml *)
module type tPILE = sig
type 'elem pile
val vide: 'elem pile
val empiler: 'elem -> 'elem pile -> 'elem pile
val depiler: 'elem pile -> 'elem pile
val sommet: 'elem pile -> 'elem
val estVide: 'elem pile -> bool
end

(* test.ml *)
open Test_lib
open Report
let () =
  set_result @@
  ast_sanity_check code_ast @@ fun () ->
  [ Section
      ([ Text "Module:" ; Code "PILE" ],
       test_module_property [%ty : (module tPILE) ] "PILE"
         (fun (module P : tPILE) ->
           (* abstract test suites built from the constructors of the ADT *)
           let suite_estVide_int = [(P.vide, fun () -> true);
                                   (P.(empiler 1 (empiler 2 vide)), fun () -> false)]
           and suite_estVide_bool = [(P.(empiler true vide), fun () -> false)]
           and suite_sommet_int = [(P.(empiler 1 (empiler 2 vide)), fun () -> 1)]
           and suite_depiler_int = [(P.(empiler 1 (empiler 2 vide)), fun () -> empiler 2 vide)] in
           (* check assertions w.r.t. the observers of the ADT *) 
           test_immediate_1 P.estVide suite_estVide_int @
           test_immediate_1 P.estVide suite_estVide_bool @
           test_immediate_1 P.sommet suite_sommet_int @
           test_immediate_1 P.depiler suite_depiler_int));
  ]
```

What do you think?
If you are OK with this, I could do more tests, then propose a PR to add such `test_immediate` facilities?

AltGr:
Yes, that's another good solution ; I had discarded it because it doesn't allow direct comparison against the solution, but since you can run tests as you would normally for top-level functions, possibly with different (explicit) type instanciations (as in your example), comparing against the solution is probably not the best approach to take here, and your example proves it's quite readable.

yurug:
@lyrm is already working on a similar feature if I understood correctly what you are about to do.

erikmd:
Hi @yurug, actually I did not perform much progress on this as I am currently in holidays, and I've just seen your [comment](https://github.com/ocaml-sf/learn-ocaml/pull/57#issuecomment-410733651).

As I wouldn't want to trigger duplicate of efforts and since this feature is not that "urgent" for us in Univ. Toulouse III (we will tackle OCaml-modules-based ADTs in the last course of the upcoming semester), maybe I can wait a little and re-discuss this feature a bit later with you and @lyrm and @AltGr...

Issue #58, opened 2018-07-11
Grade incomplete solutions?


Eemency:
I’m writing a program intended to check the validity of the tests proposed by the teachers.

The idea is to have them write several “essays”, that are more or less correct copies of students inside an `essay` directory inside each exercise directories. Each essay will be associated to the grade the teacher expects a student to have for this essay.
Then, the program would grade them to see if the expected grade matches the generated one. The goal is to detect errors in the test function or in the solution.

I’m using the `grade` functionality of the `learn-ocaml` command. Sadly, I didn’t find a way to fetch the score obtained by a graded essay (with `--grade-student`). For example, if I use `learn-ocaml grade --display-progression` in the exercise directory, I get full points as the solution is tested, which is correct. That also gets printed on the screen. Same if I grade a correct solution. But if I grade an incorrect solution, I only get “Failure” and no more informations, even though some answers are correct. The generated grade misses, so I can’t check it against the defined one.

To put it in a nutshell, I think it would be interesting to give the grade obtained when using `grade` from the command line instead of just outputting “Failure”.

AltGr:
The interface of the `grade` command is indeed quite rough (ported directly from the old `learnocaml-grader` exec) and clearly needs to be rewritten. Adding this to the to-do list.

Related to #41 in this regard.

EmileRolley:
Closed by #63

Issue #41, opened 2018-06-26
Error message of learn-ocaml grade


yurug:
The following error message is a bit disturbing. (In that example, `tmp` does not contain any learn-ocaml related material.)

```
yann➜/tmp» learn-ocaml grade                                                                                                                                                                            [15:22:00]
Fatal error: exception Learnocaml_exercise.Missing_field("solution.ml")
Called from unknown location
Called from unknown location
Called from unknown location
Called from unknown location
Called from unknown location
Called from unknown location
Called from unknown location
```

yurug:
Now we get:
```
yann➜/tmp» learn-ocaml grade                                                                                                                                                                            [10:18:35]
[ERROR] Could not load exercise at .: Learnocaml_exercise.File.Missing_file("solution.ml")
```
I think this is good enough.

PR #63, opened 2018-07-19
Improve the grader, fix errors not reported by 'build'

### LINKED ISSUE #59: No error when building exercise without an appropriate sampler
The issue is simple: an user type is both defined in `prelude.md` and in `test.ml` but no sampler for this type is given. An error should be raised when building exercise session. However, it is not the case, and so, it's resulting in a failure report when the user tries to grade the exercise. 

Obviously, user types must not be redefined in test.ml but still, an error should be raised at building time.

A dummy example: 
 **prelude.ml:**
```ocaml
type ab = A | B
````
**solution.ml :**
```ocaml
let foo x = match x with
  | A -> 0
  | B -> 1
```
**test.ml**
```ocaml 
type ab = A | B
let exercise_1 =
  Section ([ Text "Function: " ; Code "foo" ], 
           test_function_1_against_solution 
             [%ty: ab -> int] 
             "foo"                  
             ~gen:10                            
             []               
          )
```

Moreover, the teachers who report this issue actually did not have a nice failure report, but a "Division_by_zero" error but I did not manage to reproduce this behavior.

AltGr:
Closes #59, #58.

### DIFF
--- a/src/grader/build.ocp
+++ b/src/grader/build.ocp
@@ -117,6 +117,7 @@ begin library "grading-jsoo"
 end
 
 begin program "learnocaml-grader-worker"
+  comp_requires = [ "ppx_ocplib_i18n:asm" ]
   requires = [
     "toploop_jsoo"
     "grading"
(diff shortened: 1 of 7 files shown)


AltGr:
Added fix for #41.

PR #64, opened 2018-07-23
Refactoring of exercises processing

### LINKED ISSUE #26: title.txt and meta.json
Is there a good reason that explains why the title is not store in the metadata file?

OCamlPro-Couderc:
This PR refactors slightly how exercises are read and processed. It adds a proper representation for exercises, and correctly propagates the informations from `meta.json`.

The module `Learnocaml_exercise` now contains a proper representation for exercises, with its own encoding. The previous representation is still used as an intermediate state, which is still useful to store the result of reading files. As of now, the function `get` and `set` are both replaced by respectively `access` and `decipher`, and `update` and `cipher`. As their name suggest, only `cipher` and `decipher` can encode and decode files properly.

This also fixes #26

### DIFF
--- a/src/app/learnocaml_exercise_main.ml
+++ b/src/app/learnocaml_exercise_main.ml
@@ -258,6 +258,16 @@ let () =
             },\n"
          *)
        in
+       (* Looking for the description in the correct language. *)
+       let descr =
+         let lang = "" in
+         try
+           List.assoc lang (Learnocaml_exercise.(access File.descr exo))
+         with
+           Not_found ->
+             try List.assoc "" (Learnocaml_exercise.(access File.descr exo))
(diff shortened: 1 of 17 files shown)


OCamlPro-Couderc:
Issue #65 is now fixed, and the format is ready to support translations of the exercises descriptions.

OCamlPro-Couderc:
The PR has not been merged entirely. The last 3 commits are missing (which makes #67 irrelevant).

AltGr:
AH! Gotcha... I merged the `refactor-process` branch from your fork, instead of the one on `ocaml-sf`, which was the one the PR was made from. Fixing...

AltGr:
Reverted the bogus merge, and pushed a proper one. We should finally be in sync!

PR #66, opened 2018-07-24
Some QOL improvements

AltGr:
- generate the index.json automatically, if it doesn't exist, even for hierarchies of exercises
- generate formatted JSON output from the CLI client
- fix for the Dockerfile


AltGr:
Added improvements to the client, `grade` command, and better error reporting in several cases

PR #79, opened 2018-08-15
First draft for an overview tutorial for teachers. (NOT-TO-BE-MERGED-YET)

yurug:
I am trying to document what will be the basic workflow for teachers to use Learn-OCaml in the classroom. 

For the moment, there is nothing about user exercise queue management, only basic procedures to prepare and to run a course using the platform. 

This is a draft which approximatively anticipates your upcoming PR @AltGr but does it make sense?

### DIFF
--- a/docs/howto-install-learn-ocaml-client.md
+++ b/docs/howto-install-learn-ocaml-client.md
@@ -1,0 +1,24 @@
+How to install the learn-ocaml client
+=====================================
+
+This section explains how to install a learn-ocaml client for students.
+
+## Manual compilation
+
+Note: you need a working ```opam``` environment with OCaml ```4.05.0```.
+
+* Install the dependencies using:
+  - if you have opam >=2.0 available:
+``
+opam switch create . --deps-only && opam install opam-installer && eval $(opam env)
+``
+  - otherwise:
+``
+make build-deps
+``
+(You may want to first read the script `scripts/install-opam-deps.sh` to check what it does.)
+
+* Compile the learn-ocaml management tool using:
+```
+make && make opaminstall
+```
--- a/docs/howto-use-learnocaml-in-my-classroom.md
+++ b/docs/howto-use-learnocaml-in-my-classroom.md
@@ -1,0 +1,106 @@
+How to use Learn-OCaml in my classroom
+======================================
+
+You want to teach OCaml programming using Learn-OCaml? This
+documentation provides you an overview of the steps required to use
+Learn-OCaml in your classroom.
+
+## Step 1: Deploy a learn-ocaml instance on your own machine
+
+To prepare your teaching material, you need to have a working instance
+of Learn-OCaml on your machine. Follow the tutorial:
+
+[How to setup an environment to develop exercises?](https://github.com/ocaml-sf/learn-ocaml/blob/master/docs/howto-setup-exercise-development-environment.md)
+
+## Step 2: Write exercises
+
+Learn how to write your first exercises using our dedicated tutorials:
+
+[How to write exercises?](https://github.com/ocaml-sf/learn-ocaml/blob/master/docs/howto-write-exercises.md)
+
+## Step 3: Deploy the infrastructure
+
+Let us assume that you have developed the exercises for the first week
+of your semester and that you want to setup the infrastructure to check
+that your students will be able to actually work on these exercises.
+
+Ask your system administrator to do two things:
+
+1. Deploy an instance of Learn-OCaml on a server (reachable by your students).
+   Here is a documentation that your system administrator can follow:
+
+[How to deploy an instance of the Learn OCaml server?](https://github.com/ocaml-sf/learn-ocaml/blob/master/docs/howto-deploy-a-learn-ocaml-instance.md)
+
+2. If you want your students to use the command line (not the web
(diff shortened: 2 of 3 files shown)


docs/howto-install-learn-ocaml-client.md
```diff
@@ -0,0 +1,24 @@
...
+
+## Manual compilation
+
+Note: you need a working ```opam``` environment with OCaml ```4.05.0```.
+
+* Install the dependencies using:
+  - if you have opam >=2.0 available:
+``
+opam switch create . --deps-only && opam install opam-installer && eval $(opam env)
+``
```
AltGr: Note that this will recompile a 4.05.0 switch anyway, so OCaml 4.05.0 is not really needed before this point.
Also, we should probably recommend to add `--locked` to the `switch create` command, as that ensures they get the same versions of dependencies as we do (avoids e.g. the issue we had at some point with incompatible versions among the  js_of_ocaml libs)

docs/howto-install-learn-ocaml-client.md
```diff
@@ -0,0 +1,24 @@
...
+  - otherwise:
+``
+make build-deps
+``
+(You may want to first read the script `scripts/install-opam-deps.sh` to check what it does.)
```
AltGr: Err, I think the removal of this compat layer was alredy merged ^ ^
we don't need any specific pinnings anymore, so it wouldn't be too complicated to add back something simpler if compat with opam 1.2 is still needed.

docs/howto-install-learn-ocaml-client.md
```diff
@@ -0,0 +1,24 @@
...
+
+* Compile the learn-ocaml management tool using:
+```
+make && make opaminstall
+```
```
AltGr: In this case, it's probably simpler to just remove the `--deps-only` option above, so this whole section could be replaced by just:
```
opam switch create . --locked && eval $(opam env)
```

no need to call `make` manually ! :)

AltGr: Note that this applies not just to the client -- the others probably need updating.
And it's also possible to install without even cloning the source manually, e.g.:
```
opam switch create learn-ocaml 4.05.0 # optional if the user is already on a 4.05.0 switch
opam pin https://github.com/ocaml-sf/learn-ocaml.git --locked
eval $(opam env)
```

docs/howto-use-learnocaml-in-my-classroom.md
```diff
@@ -0,0 +1,106 @@
...
+of your semester and that you want to setup the infrastructure to check
+that your students will be able to actually work on these exercises.
+
+Ask your system administrator to do two things:
+
+1. Deploy an instance of Learn-OCaml on a server (reachable by your students).
+   Here is a documentation that your system administrator can follow:
+
+[How to deploy an instance of the Learn OCaml server?](https://github.com/ocaml-sf/learn-ocaml/blob/master/docs/howto-deploy-a-learn-ocaml-instance.md)
+
```
AltGr: Insert info on using https here ?
We should probably do some more tests first.

docs/howto-use-learnocaml-in-my-classroom.md
```diff
@@ -0,0 +1,106 @@
...
+## Step 5: Upload your exercise repository to the server
+
+At any time, you can upload your exercise repository to the server to
+make new exercises available or the modify existing ones. From your
+repository, simply do:
+
+```
+learn-ocaml-admin --update .
+```
```
AltGr: that will require a little work (and a bit of testing!)
at the moment:
- we don't handle uploading the content. It sounds much safer to me to do this outside of the application, actually, e.g. using scp rather than going through our HTTP server.
- at the moment, changing the exercises needs a server restart (but a "refresh" command could be added, or even answering to SIGHUP)

docs/howto-use-learnocaml-in-my-classroom.md
```diff
@@ -0,0 +1,106 @@
...
+application. The student can program inside the browser
+or write the answer in a file. This file can be submitted
+using the command line tool:
+
+```
+learn-ocaml-client --id exercise-id myanswer.ml
+```
+
+Notice that each exercise has an identifier that is displayed
+on the web application.
```
AltGr: todo, but yes, that was already reported.
Related: #94

docs/howto-use-learnocaml-in-my-classroom.md
AltGr: (note that a server restart _won't_ affect students though, since the app is client-side)

docs/howto-use-learnocaml-in-my-classroom.md
```diff
@@ -0,0 +1,106 @@
...
+### Step 7: Download students answers
+
+You can get your student grades from the web application or by:
+
+```
+learn-ocaml-admin --tag class2018 download-grades
+```
+
+This will generate a CSV file with three columns -- the student
+username, the exercise identifier and the number of points.
```
AltGr: the current format is actually:
> student token, student name, ex1 grade, ex1 last-commit-date, ex2 grade, etc.

with the grades scaled to 100 points max for each exercise
but that could easily be adjusted.

PR #85, opened 2018-08-23
Many small fixes, big reorganisation preparing for a more full-fledged server

AltGr:


AltGr:
Some more documentation on this:

- the client-server API is now entirely defined in `state/learnocaml_api.mli`. It's a GADT containing request and response types, for each query. The module is shared between client and server, and handles encoding/decoding to/from the HTTP request details (method, path, data).

- the underlying types have been gathered into a `learnocaml_data` module, with one submodule per important type, and their Json encodings, with a consistent interface

- the `learnocaml_store` module builds upon `learnocaml_data`, extending it for the server-side with functions to store and retrieve the data. It's also uses `learnocaml_index`, which only contains the file locations (shared with the `learnocaml_process_xxx` modules)

- the `learnocaml_server` module contains the actual implementation for handling the requests from `learnocaml_api`, and the actual server-running code

- the `server_caller` module has the javascript side of calling the requests in `learnocaml_api`

There is more in this PR (like actually extending the API); but I thought explaining the base design would help!

PR #88, opened 2018-08-29
Generalize assignments

yurug:
This pull request introduces a generalization of assignments. 

Currently, assignments are manually defined by the teacher by picking students one-by-one in a list widget and by assigning them an assignment. This is not convenient for at least two reasons: (i) the students list can be very long ; (ii) the students list can be incomplete because some students might have not registered at the time of homework creation.

Now, assignments are defined by a precondition (i.e. a filter predicate) which is applied automatically on each student at homework creation time and when a new student registers.

### DIFF
--- a/src/state/learnocaml_store.ml
+++ b/src/state/learnocaml_store.ml
@@ -237,7 +237,26 @@ module Token = struct
           Lwt.fail_with "token already exists"
       | e -> raise e
 
-  let create_student () = create_gen random
+  let check_for_student_assignments token = Exercise.Status.(
+    let check_exercise_status s =
+      match s.status with
+      | Open | Closed ->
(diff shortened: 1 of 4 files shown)


yurug:
This is WIP but @AltGr, can you tell me if it seems reasonable? Also, I do have a FIXME for you.

src/state/learnocaml_store.ml
```diff
@@ -233,12 +233,18 @@ module Token = struct
       | e -> raise e
 
   let check_for_student_assignments token = Exercise.Status.(
+    let student =
+      (* FIXME:@yurug: I do not know how to load the student descriptor
```
yurug: @AltGr Can you help me with this one?

PR #89, opened 2018-08-30
Metadata support in client

OCamlPro-Couderc:
Not meageable yet.

This PR adds a new tab in the exercise interface which shows the metadata. 
It shows the description of the exercise, with the required and trained skills, and the forward and backward exercises. Information about these exercises and skills are retrieved on demand using the API.

Another addition is the aggregation of trained and required skills into two distinct maps: these associate a skill with the exercises that focus on (or require) it.

For now, consider it as a proof of concept: data are displayed simply, and minimal effort has been done on CSS now (as you will see). Skill names and exercise names are clickable.

One interesting feature would be to compute some simple form of dependency graph of an exercise : which exercises should be done first according to specified backward exercises and skills required.


src/app/learnocaml_exercise_main.ml
```diff
@@ -21,7 +21,7 @@ open Learnocaml_common
 open Learnocaml_data
 
 let init_tabs, select_tab =
-  let names = [ "text" ; "toplevel" ; "report" ; "editor" ] in
+  let names = [ "text" ; "toplevel" ; "report" ; "editor"; "more" ] in
```
yurug: "more" is a bit too generic :-)

src/app/learnocaml_exercise_main.ml
```diff
@@ -93,6 +93,172 @@ let display_report exo report =
...
+    | Exercise.Meta.Problem -> pcdata [%i"problem"]
+    | Exercise.Meta.Exercise -> pcdata [%i"exercise"] ]
+
+(* Inefficient version: Reload the exercise description each time. *)
+let display_exercise_more content_id token id =
+  let open Tyxml_js.Html5 in
+  let open Learnocaml_data.Exercise in
+  let content = find_component content_id in
+  token >>= fun token ->
+  Server_caller.fetch_exercise token id >>= fun (meta, _, _) ->
```
yurug: This call can be memoized: exercise description are static for a given running instance.

src/app/learnocaml_exercise_main.ml
```diff
@@ -93,6 +93,172 @@ let display_report exo report =
...
+    | `Requirements s -> Learnocaml_api.Requiring_skill s
+  in
+  Server_caller.request_exn req >>= fun exs ->
+  Manip.replaceChildren content @@ display_list @@ List.map pcdata exs;
+  Lwt.return ()
+
+let display_more token ex_meta id =
+  let open Learnocaml_data.Exercise in
+  let open Tyxml_js.Html5 in
+  let display_skill_link content_id s =
```
yurug: The next two functions seem to share a common structure. Can't you factorize them?

src/repo/learnocaml_index.ml
```diff
@@ -32,3 +32,7 @@ let tutorial_index_path = "tutorials.json"
 let tutorials_dir = "tutorials"
 
 let tutorial_path id = tutorials_dir ^ "/" ^ id ^ ".json"
+
+let focus_path = "focus.json"
```
yurug: Can you document the purpose of these two files in the PR description, please?

OCamlPro-Couderc:
The PR is clearly not complete, for now it only contains some form of proof of concept. I plan on refactorize and think of better names before considering merging it ;-)

yurug:
Sorry, I thought you still wanted some feedback. But I now understand you want a more high-level one.

I think it is a good idea to create these pages to describe exercises and to make them available to students while they are learning: this increases, what researchers in education call cognitive clarity.

It is also important to make this description available to the teachers when they are trying to find the right exercise to assign to a student. Would that be possible to link this description to the items of the exercise list shown in the teacher page?

Also, I do not understand the exact purpose of the two new files. Can you extend the PR description to explain that?

OCamlPro-Couderc:
Sorry, I am used to create PR for work in project stuff (mainly for tracking purpose and for general features/comments), it was not really clear :-)

I agree with your points, these information are valuable for a teacher and a student, and should be integrated in their respective UI.

The two new files are mainly used as database: they associate for each skill the exercises that require it (or focus on it). This avoids to scan the entire index each time we need to compute the set of exercises to train a particular skill.

yurug:
So these two files will be computed by `learn-ocaml build`, right? (Maybe they already are computed in your PR and I missed that part.)

OCamlPro-Couderc:
Yes, exactly (https://github.com/ocaml-sf/learn-ocaml/pull/89/files#diff-c8c9cb3c3069ebec9cb911df0ade8372R201).

OCamlPro-Couderc:
It took a bit of time but I think the PR is ready. It also adds skills into the teacher tab, but only as a way to review the entire skill base. A future PR will add filtering and more importantly the ability to generate a set of exercises to train a given skill (or as requirement of an exercise). A really naive algorithm is already included into `Learnocaml_data`, but I'm currently working on it to be more precise.

I've updated the translations and names, I hope it will be clearer this way :)

OCamlPro-Couderc:
OK, so I removed the skill tree part since it is not used in this specific PR, and I'll add it with a newer version of the algorithm later. It should be mergeable :) (the "id not found" error we had earlier was due to an inconsistent opam environment).

PR #99, opened 2018-09-01
Api simplification

lyrm:
Add some API simplification for test_lib and their documentation. 

- grade_function : same as test_function except texts for report headers and `set_progress` texts are pre-built. Enables shorter graders.

- remove test_function_against: useless now since `solution.ml` need to contain every graded functions.


yurug:
@lyrm Can you rebase your branch please?

yurug:
@lyrm There seems to remain a conflict to solve.

Issue #97, opened 2018-09-01
Exercise metadata fields not recognized


adhameer:
On the latest version of learn-ocaml, certain exercise metadata fields (in `exercises/<exercise-name>/meta.json`) don't seem to be recognized. Trying to run `learn-ocaml build` when a metadata file contains one of those fields gives the error message:

```
Updating app at ./www
Unexpected object field short_description
```

The fields I've experienced problems with are `short_description`, `authors`, `forward_exercises`, and `backward_exercises`.

A full example meta.json file that is causing problems:

```
{
  "learnocaml_version" : "1",
  "kind"               : "exercise",
  "stars"              : 0,
  "title"              : "Homework 1",
  "short_description"  : "Week 1",
  "identifier"         : "hw1",
  "authors"            : [["test author", "test.author@gmail.com"]],
  "focus"              : [],
  "requirements"       : [],
  "forward_exercises"  : [],
  "backward_exercises" : []
}
```

Changing the version to 2 doesn't fix the issue.

OCamlPro-Couderc:
Indeed, that's a bug I identified recently. There's an inconsistency between the documentation and the format read by learnocaml. These fields are described as "forward" and "backward". However, "short_description" should be correct (did you indeed set version to 2?).

I'll fix it as soon as I can.

adhameer:
When changing the version to 2, I'm still getting an error about short_description.

adhameer:
After looking at [this line of code](https://github.com/ocaml-sf/learn-ocaml/blob/master/src/state/learnocaml_data.ml#L349), I tried changing it to "shortDescription" and that fixed the issue.

yurug:
@AltGr is it intended? The duplicates between `repo/learnocaml_meta.ml` and `state/learnocaml_data.ml` should be removed.

By the way, I do prefer `short_description` to `shortDescription`.

OCamlPro-Couderc:
Inconsistency fixed in #100 (and follows the format in the documentation). I did not remove `repo/learnocaml_meta.ml`.

AltGr:
Sorry, my bad for not doing proper cleanup... repo/learnocaml_meta.ml should be removed, it's superseded by learnocaml_data.ml. I removed lors of duplications between the input formats and the repository storage formats, which were almost the same, but all had their own definitions with slightly incompatible encoders.
Also, the change from `short_description` to `shortDescription` was not intentional (and due to these incompatibilities). Should I fix it ?

PR #100, opened 2018-09-03
Fix meta.json encoding

OCamlPro-Couderc:
See #97 .


yurug:
Since we are talking about metadata: what is the status of the "title" field? If it absent and `title.txt` is present, is the title defined? I encountered a problem on a repository which is using the version 1 of meta.json: even though `title.txt` was present, the title field was undefined.

OCamlPro-Couderc:
I fixed this issue a while ago, @lyrm told me about it. I just checked, it works for me.

yurug:
OK. I will check that later.

PR #101, opened 2018-09-03
Allow creating open assignments

AltGr:
that will affect any newly created students.


AltGr:
Supersedes #88 (although I didn't keep its most advanced patches, allowing preconditions for automatically enabling assignments)

PR #111, opened 2018-09-04
Add server-side merge of assignment changes

AltGr:


src/state/learnocaml_data.ml
```diff
@@ -444,6 +444,52 @@ module Exercise = struct
       in
       by_status_explicit base a |> STM.bindings
 
+    let three_way_merge ~ancestor ~theirs ~ours =
```
yurug: I am not sure to understand this function: can the merge encounter conflicts? Can it fail?

src/state/learnocaml_data.mli
```diff
@@ -178,6 +178,12 @@ module Exercise: sig
     val by_status:
       Token.Set.t -> assignments -> (status * Token.Set.t) list
 
+    (* Merges all changes from [theirs] and [ours], based on [ancestor]. [ours]
```
yurug: By privileging "ours", you choose a last-arrived-get-the-prize strategy right?

AltGr: yes, as mentionned in the commit notes. This happens for conflicting changes on same exercise + same student.
If we choose the alternative (refuse conflicting changes), we need to add a way to report the partial error to the client.

yurug: That's the good design choice I think. I just wanted to check I correctly understood.

PR #112, opened 2018-09-04
Force entering a nickname in the clients

AltGr:


src/app/learnocaml_index_main.ml
```diff
@@ -1572,15 +1572,17 @@ let init_token_dialog () =
...
-          Learnocaml_local_storage.(store nickname) n;
-          let nickname_field = find_component "learnocaml-nickname" in
-          (Tyxml_js.To_dom.of_input nickname_field)##.value := Js.string n;
-          Some n
-    in
-    Server_caller.request_exn (Learnocaml_api.Create_token (None, nickname))
+    let nickname = Manip.value input_nick in
+    if String.length nickname < 2 then
+      (alert ~title:[%i"INVALID NICKNAME"]
+         [%i"You must provide a nickname"];
```
yurug: A nickname longer than 2 characters? We should also make sure that it contains a significant number of non-layout characters if that is possible.

PR #118, opened 2018-09-05
Teacher interface: add pseudo-checkboxes

AltGr:
these add a bit of screen noise, but probably make the interface easier to
understand.


AltGr:
Here is what it looks like:
spectacle i30796

yurug:
To me, this is indeed better with the checkboxes. What do you think @adhameer?

adhameer:
It definitely looks better to me.

PR #120, opened 2018-09-06
Fix error with 'sync' on expired exercises

### LINKED ISSUE #116: Internal error when syncing after assignment deadline
When working on an exercise after the assignment deadline has passed, clicking the Sync button results in an Internal Error (Not Found). It would be nice to have something more user-friendly.

AltGr:
Closes #116

the error actually happened only if there was no submission before the 
deadline, and with the 'sync' button, but not the 'grade' button (which also
does sync after doing the grading).

### DIFF
--- a/src/app/learnocaml_common.ml
+++ b/src/app/learnocaml_common.ml
@@ -428,7 +428,6 @@ let countdown ?(ontimeout = fun () -> ()) container t =
   in
   let rec callback () =
     let remaining = int_of_float (deadline -. gettimeofday ()) in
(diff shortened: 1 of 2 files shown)


AltGr:
I actually have doubts about the "sync" button: we should probably separate "submitted solution" and "saved solution".
The reason is that if the sudent grades, then edits, then syncs, then the submitted report and solution may no longer match, which I think could be a problem.

AltGr:
(a quick fix could be to include the supplied solution within the report!)

yurug:
Your quick fix seems fine to me. If the solution has been updated, this change has been made by the very same student: hence, updating the local solution seems the right thing to do. I imagine a student that has opened a tab on his machine at home and works during the day on another machine: when he comes back home, the solution should be updated.

PR #122, opened 2018-09-06
Separate submitted solutions from saved state, sync just the current exercise

AltGr:
This makes a few changes in the "sync" mechanism, separating it from
solution submitting (even if the same API call is used).

- solutions are only submitted on using the `Grade!` button: no need to
 repeat sending them (this also defeats the 'editing my own grade'
 hack, although without any safety since it's just client-side)

- the rest of the workspace is saved as before; but the server stores
 the editor contents separately from what was used for the submissions

- it seemed a good idea to actually avoid submitting the report after
 grading when the grade went down. Are there any downsides to that ?

- students see only:
  - their latest editor contents (not what was submitted)
  - while working, obviously, their latest report
  - on reconnecting, the report that was submitted (i.e. the best one
   that was submitted before the deadline)

This patch is backwards-compatible with the storage formats, but adds new
fields. It works but more testing would be better.

### DIFF
--- a/src/app/learnocaml_common.ml
+++ b/src/app/learnocaml_common.ml
@@ -379,16 +379,28 @@ let set_state_from_save_file ?token save =
   let open Learnocaml_local_storage in
   match token with None -> () | Some t -> store sync_token t;
   store nickname save.nickname;
-  store all_exercise_states save.all_exercise_states;
+  store all_exercise_states
+    (SMap.mapi (fun id ans ->
+         match SMap.find_opt id save.all_exercise_editors with
+         | Some (mtime, solution) ->
+             {ans with Answer.solution; mtime}
+         | None -> ans)
+        save.all_exercise_states);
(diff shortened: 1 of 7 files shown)


yurug:
> it seemed a good idea to actually avoid submitting the report after
> grading when the grade went down. Are there any downsides to that ?

I do not understand. Are you saying that the server only keeps the best answer?

AltGr:
yes, basically, now answers are only submitted on pushing the "Grade" button _and_ if the grade is at least as good as the already registered submission.

The contents of the editor are saved separately, though.

yurug:
I am not sure that we should submit answers only if grade are better. Imagine a case where the grader is not complete enough and give all the points to an answer that is not perfect: even though a student is able to finally refine his answer to reach this perfect answer, he will never be able to submit this answer to the server.

AltGr:
I can easily withdraw that part, in that case. Note that a new answer _with the same grade_ is still submitted :)

PR #125, opened 2018-09-07
Allow custom MathJax configs

wyschean:
This PR allows exercise authors to set their own MathJax config options. For example, enabling the `asciimath2jax.js` preprocessor to reduce the number of `<script>` tags can be done by adding
```html
<script type="text/x-mathjax-config">
  MathJax.Hub.Config({
      extensions: ["asciimath2jax.js"],
      asciimath2jax: {
          ...
      }
  })
</script>
```
to the beginning of the exercise description.

- MathJax startup is delayed until after the exercise description, as described in the [docs](https://docs.mathjax.org/en/latest/configuration.html#delaystartupuntil).
- Removes `elements: ["learnocaml-exo-tab-text"]` from the hardcoded config. Since the exercises are loaded as an iframe, their tags don't inherit the `learnocaml-exo-tab-text` class, so all the line does is in some cases prevent extensions from working properly.

### DIFF
--- a/src/app/learnocaml_exercise_main.ml
+++ b/src/app/learnocaml_exercise_main.ml
@@ -276,7 +276,6 @@ let () =
          "MathJax.Hub.Config({\n\
          \  jax: [\"input/AsciiMath\", \"output/HTML-CSS\"],\n\
          \  extensions: [],\n\
-         \  elements: [\"learnocaml-exo-tab-text\"],\n\
          \  showMathMenu: false,\n\
          \  showMathMenuMSIE: false,\n\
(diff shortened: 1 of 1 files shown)


AltGr:
Thanks! Indeed Mathjax is loaded only on the specific iframe anyway, so the `elements:` config wasn't needed. Among our plans is to allow writing exercises in Markdown, with an easier syntax for math blocks — like we already do for tutorials in TryOCaml — with which the `asciimath2jax` preprocessor would not be needed. This doesn't contradict this at all though :)

Issue #123, opened 2018-09-06
Customizing the "nickname" field


adhameer:
We are using the Learn OCaml platform in a course where we need students to use their student ID number as their nickname so that we can associate their grades to them. Since there is no explanation of the nickname field on the welcome page, several have written something else in the nickname field instead. I imagine this would be a problem in any course where Learn OCaml is used.

It would be nice to have some customization of the welcome screen - at least being able to add a message at the top, or ideally to change the nickname field to say "student ID" or something similar.

yurug:
We could implement that kind of customization by using as label the content of a file named "welcome-nickname-label.txt" at the root of the folder. If it is not present, the default label would appear.

AltGr:
I have patches for this but didn't have time to file a PR yet :)

yurug:
We need to make the configuration file persist through `learn-ocaml build`.

AltGr:
Indeed!

PR #130, opened 2018-09-11
Allows some basic configuration of the welcome screen messages

AltGr:


yurug:
I do not understand how the user interface to this configuration system. Can you explain how this works please?

AltGr:
there is not much at the moment, that part needs improvement: one would need to first run `learn-ocaml build`, then edit the `www/js/learnocaml-config.js` by hand.

The idea is that options should be added to `learn-ocaml build` to allow specifying this kind of things. Note that this is already how the `enable|disable` flags for the different tabs are handled (with appropriate options, in the this case)

yurug:
Thanks for the explanations. 

Does `learn-ocaml build` replace the configuration file each time it is run?

Should we document this?

AltGr:
> Does `learn-ocaml build` replace the configuration file each time it is run?

yes :/

> Should we document this?

if we add options to `learn-ocaml build` — or better (?), a way to supply the config file — the doc can easily be part of the man-page.

yurug:
Could we say that if a file named `learnocaml-config.js` is present in the root of the repository, then it is copied in `www/js` by the `learn-ocaml build` command?

AltGr:
In any case, there should be no downside to merging this now, and improving interface/doc afterwards. Let's just keep #123 open to make sure it doesn't stay unfinished.

AltGr:
> Could we say that if a file named learnocaml-config.js is present in the root of the repository, then it is copied in www/js by the learn-ocaml build command?

One downside of this approach is that, on errors in the file, nothing will be reported at build-time but the resulting app may be broken. I'll define a proper Json format, will take a few minutes and also help with documentation, then export it to the `learnocaml-config.js` file.

yurug:
Good idea!

PR #131, opened 2018-09-11
Server: add a server-side cache

AltGr:
avoiding to decode & recode everything from & to json, and more importantly,
avoiding recompressions.

The cache is purely in memory at the moment, but given the size of the static
files we serve, that should be no problem.

### DIFF
--- a/learn-ocaml.opam
+++ b/learn-ocaml.opam
@@ -24,7 +24,7 @@ depends: [
   "js_of_ocaml-compiler"
   "js_of_ocaml-lwt"
   "js_of_ocaml-tyxml"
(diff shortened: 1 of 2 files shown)


AltGr:
Ah, it seems that the `lwt` version used at this point didn't have `Lwt_preemptive` yet, so this should wait for my PR that upgrades the lwt version used (then, this should be rebased with  a lower bound added).

I use `Lwt_preemptive` to do the compression in a separate thread, to avoid the server becoming unresponsive when compressing large files. Of course it's just a tiny part of this PR and could be removed (it's not really related to the caching mechanism, which actually makes it a little bit less of a problem).

AltGr:
rebased

PR #135, opened 2018-09-12
Add support for Markdown in exercises

Niols:
This PR is a small modification of the `learnocaml_exercise.ml` file that allows for support for Markdown in exercises description. It modifies the functions `read_descr` and `read_descrs` in such a way that:

- `read_descr` now has a new argument that is the list of extensions to try in order. I also fixed a bug that whould have happened if `land` was not empty in `read_descrs`.

- `read_descrs` now has a list of extensions to try that it can give to `read_descr`. I believe this list should at some point move somewhere else. Maybe in the `descr` record?

I had to add a small `markdown_to_html` function. I used the basic functions from Omd, which means that it might not support other features that you want to have in your flavour of Markdown. I believe there should be a module (`Markdown`?) that implements this flavour with an interface containing at least `from_string` and `to_html`.

### DIFF
--- a/src/repo/build.ocp
+++ b/src/repo/build.ocp
@@ -7,6 +7,7 @@ begin library "learnocaml-repository"
   requires = [
     "ocplib-json-typed"
     "xor"
+    "omd"
     "lwt"
     "ezjsonm"
   ]
--- a/src/repo/learnocaml_exercise.ml
+++ b/src/repo/learnocaml_exercise.ml
@@ -278,9 +278,15 @@ module File = struct
               descrs := (lang, f raw) :: !descrs;
               return ()
       in
+      let markdown_to_html md =
+        Omd.(md |> of_string |> to_html)
+      in
       let read_descrs () =
         let langs = [] in
-        let exts = (Filename.extension descr.key, fun h -> h) :: [] in
+        let exts = [
+            (Filename.extension descr.key, fun h -> h) ;
+            (".md", markdown_to_html)
+          ] in
         join (read_descr None exts :: List.map (fun l -> read_descr (Some l) exts) langs)
         >>= fun () ->
         ex := set descr
--- a/src/repo/learnocaml_exercise.ml
+++ b/src/repo/learnocaml_exercise.ml
@@ -251,31 +251,42 @@ module File = struct
        *       with _ -> return ()
        * in *)
       let descrs = ref [] in
-      let read_descr lang =
-        begin
-          match lang with
-            None -> read_field descr.key
-          | Some lang ->
-              let key =
-                Filename.remove_extension descr.key
-                ^ "." ^ lang ^ "." ^
-                Filename.extension descr.key in
-              read_field key
-        end >>= function
-          None -> return ()
-        | Some raw ->
-
-            descrs := (lang, raw) :: !descrs;
-            return ()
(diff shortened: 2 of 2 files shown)


src/repo/learnocaml_exercise.ml
```diff
@@ -251,31 +251,48 @@ module File = struct
...
+              (* If it does not work, we go on and try other extensions. *)
+              read_descr lang other_exts
+           | Some raw ->
+              (* If it does, we apply the function, add the
+                 description to [!descrs] and return. *)
+              descrs := (lang, f raw) :: !descrs;
+              return ()
+      in
+      let markdown_to_html md =
+        Omd.(md |> of_string |> to_html)
```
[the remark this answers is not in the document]
yurug: Yes. Do it if you want.

src/repo/build.ocp
```diff
@@ -7,6 +7,7 @@ begin library "learnocaml-repository"
   requires = [
     "ocplib-json-typed"
     "xor"
+    "omd"
```
yurug: I am surprised that this dependency was not already here since `src/repo/learnocaml_parser.ml` uses `Omd_backend`. I must have missed something...

yurug:
This PR is fine to me. @OCamlPro-Couderc I let you review it since you are the expert of this piece of code.

src/repo/build.ocp
Niols: It's because the compilation of `learnocaml_tutorial_parser` is handled in an other stanza, line 16. And this stanza had this dependency specified.

yurug: OK, thanks.

AltGr:
This is great, thanks! Did'nt have time to get around to it yet :)
It would be best to plug in the markdown stuff that we use for  the tutorials: it has, for example, support for mathjax snippets. I don't know how easy it would be, though.

Niols:
> It would be best to plug in the markdown stuff that we use for the tutorials: it has, for example, support for mathjax snippets. I don't know how easy it would be, though.

@AltGr I do agree. There are a lot of things that could be extracted from https://github.com/ocaml-sf/learn-ocaml/blob/master/src/repo/learnocaml_tutorial_parser.ml. But the `parse_tutorial` functions are quite long and I don't pretend to understand exactly what everything does. It has to do with the `Tutorial.text` type found in https://github.com/ocaml-sf/learn-ocaml/blob/master/src/state/learnocaml_data.ml. This type and the functions converting from and to HTML and Markdown should probably have their own separate file (`src/utils/richText.ml`?).

> I was just thinking at one thing: we use "descr.html" as the key to represent the value `descr`. However, with the addition of internationalized description files and the support for markdown, this key is no longer relevant. It works for now since we convert everything into HTML and with one language by instance, but that's something we could improve in the future.

@OCamlPro-Couderc I do agree too. The type `'a file` is not generic enough to handle the descr file. I'm not sure if one should generalize it or agree that descr is a particular file.

Niols:
There was a problem with the merge. It's because the PR was anterior to the "dunification" of learn-ocaml. And the Omd dependency lacks in this PR. One has to change https://github.com/ocaml-sf/learn-ocaml/blob/master/src/repo/dune#L6 and add "omd" as a dependency.

PR #151, opened 2018-09-18
Use a native stand-alone server

AltGr:
This compiles the server separately from the main app. By not embedding a
grader, we can compile it to native code, giving a huge performance boost.

The learn-ocaml main exec will also find an installed native server and 
bootstrap to it, so everything is completely transparent to the user.


AltGr:
No, it shouldn't, there is just one more binary installed, everything else should be transparent

yurug:
Great! Thanks!

PR #153, opened 2018-09-19
Save editor contents locally upon leaving an exercise

### LINKED ISSUE #144: Save the source code when leaving the exercise
It the student goes back to the exercise list from the editor view, all their edits since the last time they hit "grade" are lost.

Several students encountered this problem, and lost all their edits for the exercise.

It would be nice if the state of the editor was saved.

AltGr:
Closes #144, thanks for the suggestion.


AltGr:
oops, duplicate

PR #164, opened 2018-09-27
Logout dialog + CSS fixes

### LINKED ISSUE #161: Bigger Logout button
The current "Logout" button is small and hard to find, especially compared to the size of the other buttons. It would be nice to make it more visible.

### LINKED ISSUE #163: CSS issue in teacher activity
After the students are loaded in the student pane, the students heading overlaps the search box.

image

OCamlPro-Couderc:
Improves logout by adding a dialog box asking for confirmation and displays the token as reminder for future login (fixes #161).

Also fixes #163.

### DIFF
--- a/src/app/learnocaml_index_main.ml
+++ b/src/app/learnocaml_index_main.ml
@@ -727,10 +727,10 @@ let set_string_translations () =
     "txt_returning", [%i"Returning user"];
     "txt_returning_dialog", [%i"Enter your token"];
     "txt_login_returning",  [%i"Connect"];
-    "txt-logout", [%i "Logout"];
-    "txt-logout-token", [%i "Before log out, be sure to save your token"];
-    "txt-logout-ok", [%i "Logout"];
-    "txt-logout-cancel", [%i "Cancel"];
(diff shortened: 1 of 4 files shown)


src/app/learnocaml_index_main.ml
```diff
@@ -706,6 +727,10 @@ let set_string_translations () =
     "txt_returning", [%i"Returning user"];
     "txt_returning_dialog", [%i"Enter your token"];
     "txt_login_returning",  [%i"Connect"];
+    "txt-logout", [%i "Logout"];
```
yurug: Why do you use "-" instead of "_" in the identifiers here?

OCamlPro-Couderc: Well, I suppose we should use "_" for identifiers of internationalized strings and and "-" for the rest? As far as I remember, this was not really consistent across the entire codebase actually (sometimes classes use one or the other, and ids too). Anyway, I'll change these ones ;)

yurug: Thanks! Let's not propagate chaos :)

PR #166, opened 2018-10-03
Warnings fixes

AltGr:
(and a few "silent bug" fixes along the way)

### DIFF
--- a/src/ace-lib/ace.ml
+++ b/src/ace-lib/ace.ml
@@ -101,7 +101,7 @@ let set_mode {editor} name =
 
 type mark_type = Error | Warning | Message
 
-let string_of_make_type = function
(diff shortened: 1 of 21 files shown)


src/ace-lib/ace.ml
```diff
@@ -283,4 +283,4 @@ let delete doc range =
   doc##(replace range (Js.string ""))
 
 let remove { editor } dir =
-  editor##(remove (Js.string "left"))
+  editor##(remove (Js.string dir))
```
yurug: Is it the "silent bug"?

AltGr: yes, one of them. The more dangerous one is the closing of file descriptors in the toplevel workers, which actually could close the wrong one.

PR #167, opened 2018-10-03
Remove placeholder text when there is no exercise description

Armael:
This removes the "No description available" placeholder description in the index of exercises.

This makes the index of exercises much less visually cluttered when there are exercises without a description.


yurug:
This seems fine to me. @AltGr, what do you think?

PR #170, opened 2018-10-08
Rehaul of the left panel interface

AltGr:


AltGr:
Here is what it looks like:
image

PR #171, opened 2018-10-09
Fix crash of the indentation engine

AltGr:


src/ace-lib/ocaml_mode.ml
```diff
@@ -275,7 +275,7 @@ let get_line_tokens line st row doc =
     let open Approx_tokens in
     let open Nstream in
     match Nstream.next_full stream with
-    | None | Some ({token = EOF ; _}, _, _) -> assert false
+    | None | Some ({token = EOF ; _}, _, _) -> st, List.rev tokens
```
yurug: Do you know why this executation path was marked as dead with `assert false`?

AltGr: Yes: it was a supposed invariant of the [indentation lexer](https://github.com/OCamlPro/ocp-indent/blob/nlfork/src/approx_lexer.mll).  but it turns out there are some cases where it wasn't respected. Namely, a `# directive` not properly terminated.

PR #176, opened 2018-10-14
Make code in descriptions more readable

### LINKED ISSUE #175: Blocks of code are unreadable
There is a problem with the borders of the `code` elements for multiline blocks of code (see screenshot).
The simple solution is to remove [this line](https://github.com/ocaml-sf/learn-ocaml/blob/master/static/css/learnocaml_standalone_description.css#L9) I think.

code moche

ineol:
Fixes #175 by removing borders of <code> elements.


yurug:
Fine to me. @AltGr, what about you?

PR #179, opened 2018-10-15
Rework the Stack overflow message

Armael:
The current message that is printed when a stack overflow occurs is "Your code did too many recursions.". This is wrong in some cases. 
The cause for the Stack overflow might also be that the code is not tail-recursive -- the current message is particularly confusing in a practical session which is *teaching* tail-recursion.

The PR contains a mild refactoring of the message. I kept the "too many recursions" part, but as a hint after printing the root cause (similarly to the timeout message just after).


yurug:
I guess you should update translations too.

Armael:
`git grep` seems to indicate that these messages are not translated...

yurug:
OK, then internationalisation of this message will be for another PR.

PR #180, opened 2018-10-16
Storage and retrieval of save histories using git

AltGr:
This works but probably needs some more work:
- [X] a global write mutex is used, this should be refined for each separate git repo
- [x] git needs to be added to the dependencies, esp. in the Docker images
- [X] this might be an opportunity to re-examine the save format (one-line json files)

### DIFF
--- a/src/state/learnocaml_api.ml
+++ b/src/state/learnocaml_api.ml
@@ -32,6 +32,8 @@ type _ request =
       'a token -> Save.t request
   | Update_save:
       'a token * Save.t -> Save.t request
+  | Git: 'a token * string list -> string request
+
   | Students_list:
       teacher token -> Student.t list request
   | Set_students_list:
@@ -103,6 +105,7 @@ module Conversions (Json: JSON_CODEC) = struct
           json Save.enc
       | Update_save _ ->
           json Save.enc
+      | Git _ -> str
       | Students_list _ ->
           json (J.list Student.enc)
       | Set_students_list _ ->
@@ -166,6 +169,8 @@ module Conversions (Json: JSON_CODEC) = struct
         get ~token ["save.json"]
     | Update_save (token, save) ->
         post ~token ["sync"] (Json.encode Save.enc save)
+    | Git _ ->
+        assert false (* Reserved for the [git] client *)
 
     | Students_list token ->
         assert (Token.is_teacher token);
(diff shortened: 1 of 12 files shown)


yurug:
Great! Does that mean that the API can be extended to allow downloading the entire git repository of a user?

About the first two items : they seem important enough to be integrated into this PR before merge.

About the third one, I guess there are two possibilites:
- indented json
- YAML

AltGr:
> [can] the API be extended to allow downloading the entire git repository of a user?

I guess it could, we'd just need to make sure we let the HTTP calls used by `git` to pass through our server.
That would be much more convenient if we saved separate .ml files rather than a single .json file, though.
One possibility would be to have different content in the save .json file (all the info, but only the latest version, as it is now), and the git tree (containing the extracted .ml files separately, with their history). A bit more involved, but could be convenient.
An hybrid solution would be to do just as now, but _add_ the expanded source files to a git tree. That would also be easier to implement.

> they seem important enough to be integrated into this PR before merge.

Yes, that was the intent :)

> About the third one, I guess there are two possibilites: [...]

we just have one option to turn on for indented json :). That would probably already be better, indeed.

yurug:
Great! Does that mean that the API can be extended to allow downloading the entire git repository of a user?

About the first two items : they seem important enough to be integrated into this PR before merge.

About the third one, I guess there are two possibilites:
- indented json
- YAML

yurug:
> > [can] the API be extended to allow downloading the entire git repository of a user?
> 
> An hybrid solution would be to do just as now, but _add_ the expanded source files to a git tree. That would also be easier to implement.

If I understand correctly there would be some redundancy between the JSON file and this git tree, right? This is probably insignificant.

That seems a good solution. Let's do that.

yurug:
Ping @AltGr

AltGr:
Yup, getting back to this, the utf8 issue took me some time (had to dig all the way to `js_of_ocaml`)

AltGr:
This last update should make it fairly complete:
- git storage is using mutexes properly
- solutions are stored as `EXERCISE_ID.ml` files (using subdirs if the exercise IDs contain slashes) for each student. The grades are included in comments for convenience.
- git repositories can be retrieved with `git clone 'BASEURL/STU-DEN-TTO-KEN/learnocaml-workspace.git'`.  Access is read-only, of course

IMPORTANT NOTE: this changes the `sync` layout a bit (automatically and lazily) in a way that is not backwards-compatible (saves files are moved from `STU/DEN/TTO/KEN` to `STU/DEN/TTO/KEN/save.json`, which leaves room for the per-student git repositories).

Other sync files (e.g. definitions of assignments) are also backed up using git, because it's quite cheap and could allow to recover after disasters. There is no API to retrieve them though.

src/state/learnocaml_api.ml
```diff
@@ -166,6 +169,8 @@ module Conversions (Json: JSON_CODEC) = struct
         get ~token ["save.json"]
     | Update_save (token, save) ->
         post ~token ["sync"] (Json.encode Save.enc save)
+    | Git _ ->
+        assert false (* Reserved for the [git] client *)
```
yurug: Not sure to understand the comment here.

AltGr: This is the function that constructs requests, and Learn-ocaml itself won't ever make this request, only receive it from the `git` command-line client.

yurug: Oh I see! Thanks!

PR #196, opened 2018-11-07
Rehaul of error handling (in particular for network errors)

AltGr:


yurug:
Can you sum up how you redesigned error handling? This is difficult to infer from the patch.

AltGr:
The third patch sums it up a bit. Basically, XHR are wrapped into a function that catches the failure and displays a dialog offering to retry, an also abort (cancelling the associated lwt thread). Before, the exception was only caught by the top-level handler, which is the infamous `INTERNAL ERROR` box which blocks everything until reload.

The case of web-workers was much more tricky, see patch notes.

AltGr:
Most of all, this fixes the issue where, in exercises, you couldn't work or grade anymore if you got disconnected from the server, because the autosave couldn't be done.

PR #198, opened 2018-11-08
Exercises dependency graph

OCamlPro-Couderc:
Generates a graph of dependencies between the exercises of a repository.
An exercise depends on:
* Its declared backward exercises
* The exercises that focusing on its required skills

This PR only includes the data structure and the graph generation, with a CLI option to dump the graph into its dot representation when building the repository. It should be useful to generate a set of exercises ordered by dependencies (a small algorithm is given as `compute_exercise_set`, but it is really naive).

For example, here is a generated repository with only 10 exercises and 3 skills (it should be clearer for a well-thought repository): 
image

### DIFF
--- a/src/repo/learnocaml_process_exercise_repository.ml
+++ b/src/repo/learnocaml_process_exercise_repository.ml
@@ -60,6 +60,14 @@ let dump_outputs = ref None
 
 let dump_reports = ref None
 
+let dump_dot exs =
+  match !Grader_cli.dump_dot with
+    None -> Lwt.return ()
+  | Some filename ->
+      let graph = Exercise.Graph.compute_graph ~filters:[] exs in
+      Lwt_io.with_file ~mode:Lwt_io.Output filename
+        (fun oc -> Lwt_io.write oc (Format.asprintf "%a" Exercise.Graph.dump_dot graph))
+
 let n_processes = ref 1
 
 let print_grader_error exercise = function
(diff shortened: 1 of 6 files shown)


src/repo/learnocaml_process_exercise_repository.ml
```diff
@@ -60,6 +60,14 @@ let dump_outputs = ref None
 
 let dump_reports = ref None
 
+let dump_dot exs =
+  match !Grader_cli.dump_dot with
+    None -> Lwt.return ()
```
yurug: Why isn't it an error to try to execute `dump_dot` if `Grader_cli.dump_dot` is undefined?

src/repo/learnocaml_process_exercise_repository.ml
```diff
@@ -201,6 +209,7 @@ let main dest_dir =
        in
        fill_structure SMap.empty structure >>= fun (all_exercises, index) ->
        to_file Index.enc (dest_dir / Learnocaml_index.exercise_index_path) index >>= fun () ->
+       dump_dot index >>= fun () ->
```
yurug: Ah OK! Please rename `dump_dot` into `dump_dot_if_asked`.

AltGr:
This could also be done client-side, with a button `download exercise graph` under "actions" in the teacher screen ? Or do we even have an easy way for web-display of .dot files ? The bonus is that it could handle "skills" defined from the web-interface.

Otherwise, maybe it would be best to keep the code outside of the client-reachable `learnocaml_data` module ?

yurug:
There exists visualizers for dot files in JS, see https://github.com/mdaines/viz.js.

PR #200, opened 2018-11-14
Fix student index with the new save file layout

AltGr:


AltGr:
Oops, my last PR introduced a bug in the teacher screen (student index). This fixes it.

PR #206, opened 2018-12-06
Synchronize dependencies in the two opam files.

yurug:
I am not sure to understand why we have two opam files but it seems reasonable to find the same dependencies declaration in both.


AltGr:
The `learn-ocaml.opam` file should already have all the required dependencies. The `.locked` file should only be generated from it by use of `opam lock`, and has overly specific dependencies, in order to allow for precise development environment reproduction.

For example, the `react` dependency is not needed because it's obtained through `lwt_react`. But the `.locked` file explicits it and fixes its version.
An exception is `conf-libev`, which has been forced by hand into the `.locked` file, since it has such a huge effect on the generated binaries.

PR #208, opened 2018-12-14
A little improvement in the setup instructions.

fpottier:
The name of the Docker image is still incorrect, though:
there is no "latest" image. Should this image be created?
Currently, then the user must add a version number such
as "0.9" to the image name.

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,7 +17,7 @@ The exercises and lessons are in another github repository:
 Howtos
 ------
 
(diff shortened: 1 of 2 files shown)


yurug:
`dev` is the correct tag and should be used instead of `latest`.

fpottier:
OK, added the tag `dev` in the instructions.

yurug:
Thank you!

PR #218, opened 2018-12-19
Fix dune install

### LINKED ISSUE #217: Dune-based installation is broken
Under a fresh cloned learn-ocaml source tree, the following sequence of commands *does not* install all the files listed in `static/dune`:

```bash
make
make install
```
while the following sequence does install all the files of `static`:
```bash
make
make install
make
make install
```

This behavior is reproducible using dune 1.1.1 and dune 1.6.2.

yurug:
We make sure that the `static` directory is built before `make build` is run.

This fixes #217.


yurug:
Sorry, bad move.

PR #220, opened 2018-12-19
Minor fixes/clarifications to AST checker documentation

adhameer:
- Specify that the function given to `on_variable_occurrence` is only
  called for uses of *free* variables (relative to the AST). I think
  this is probably true for `on_module_occurrence` as well, but I
  haven't actually looked into that.
- Fix a few examples of ill-formed arguments to `ast_check_expr`
  (namely, labelled arguments not enclosed in parentheses).
- Fix a few typos.


yurug:
Thanks!

PR #222, opened 2018-12-27
Expose Test_lib.run_timeout.

fpottier:
This function is slightly lower-level than `Test_lib.exec`, as it does not catch exceptions (and does not record `stdout` and `stderr`). The benefit is that this allows the caller to display a backtrace.


yurug:
Thank you @fpottier!

PR #229, opened 2019-01-20
Fix 0pt success score colour in failure sections

adhameer:
Currently, if a section has a total score of 0, all of the scores
within it are displayed in red font, even for 0pt success
messages. This reorders the CSS rules so that the success score
rule takes priority.

Old:
image

New:
image


AltGr:
Thanks!

Issue #244, opened 2019-05-20
Conflicting cohttp-lwt versions in locked opam file


nobrakal:
Hi,

Following the README, I tried to install the required ocaml environment for learn-ocaml and I ran (after a fresh clone on an up-to-date archlinux):
```
opam switch create . --deps-only --locked
```

Which failed and gave me:
```
[ERROR] The following local packages don't appear to be installable:
          - learn-ocaml.0.10

Do you want to create an empty switch regardless? [Y/n] y
The following dependencies couldn't be met:
  - learn-ocaml → cohttp-lwt = 1.1.1
  - learn-ocaml → cohttp-lwt-unix = 1.2.0 → cohttp-lwt = 1.2.0
Your request can't be satisfied:
  - No available version of cohttp-lwt satisfies the constraints
```

There is indeed a strange version constraint in `learn-ocaml.opam.locked`:

```
  "cohttp" {= "1.1.1"}
  "cohttp-lwt" {= "1.1.1"}
  "cohttp-lwt-unix" {= "1.2.0"}
```

Note that the two last Travis builds (from 462a8e810d0d533f8751198188201d29c0a0c014 and  aa8a36862d96aefdd605b5cc08b1dc0452319630) are failing because of this issue.
The constraint was introduced by 40ef7d0837e39c1618e2acf6ca666c474f14b074 which, for a reason that seems obscure to me, built successfully on Travis.

AltGr:
Constraints of consistency between the versions of the `cohttp-*` packages have probably been added to the opam-repository in the meantime, which would explain that this issue is new: the `locked` file is generated and I did have these packages installed together at the time.

Thanks for the PR, that should be enough to fix it!

PR #245, opened 2019-05-20
Set version of cohttp-lwt-unix from 1.2.0 to 1.1.1

Asereme:
This change fix the issue #244 , the version of cohttp-lwt-unix must be the same as cohttp-lwt.


AltGr:
Thanks!

PR #247, opened 2019-05-20
If Lwt_utils.copy_tree with empty source directory, then Lwt.return_unit 

### LINKED ISSUE #228: copy_tree fails if source directory is empty
The function `Lwt_utils.copy_tree` fails if its source directory is empty. Indeed, in that case, it invokes `cp` with no source files, and `cp` complains that it is used incorrectly.

As a result, `learn-ocaml build` fails if an exercise directory happens to contain an empty subdirectory.

Asereme:
This will avoid using cp if the directory is empty, and just return Lwt.return_unit for fix #228

### DIFF
--- a/src/utils/lwt_utils.ml
+++ b/src/utils/lwt_utils.ml
@@ -21,20 +21,24 @@ let rec mkdir_p ?(perm=0o755) dir =
       Lwt_unix.mkdir dir perm
 
 let copy_tree src dst =
-  Lwt.catch (fun () ->
-      mkdir_p dst >>= fun () ->
-      let cmd =
-        Array.concat
-          [[|"cp"; "-PR"|];
-           Array.map (Filename.concat src) (Sys.readdir src);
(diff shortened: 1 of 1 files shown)


src/utils/lwt_utils.ml
```diff
@@ -21,20 +21,24 @@ let rec mkdir_p ?(perm=0o755) dir =
...
-           [|dst|]]
-      in
-      Lwt_process.exec ("", cmd) >>= fun r ->
-      if r <> Unix.WEXITED 0 then Lwt.fail_with "copy_tree"
-      else Lwt.return_unit)
-    (function
-      | Sys_error _ | Unix.Unix_error _ -> Lwt.fail_with "copy_tree"
-      | e -> raise e)
+  let files = Sys.readdir src
+  in
```
yurug: Please move this `in` at the end of the previous line.

Issue #214, opened 2018-12-18
Slightly inconsistent font sizes in Markdown exercise


fpottier:
When an exercise is written in Markdown (in a file `descr.md`), verbatim text (between backquotes) appears in a larger font than normal text. Ideally all text should be in the same font size.

As a side note, it would be nice if the default font size was a bit larger.

It would also be nice if the user could control the font size.

PR #248, opened 2019-05-21
Enlarge body font size

nobrakal:
Hi,

This fix a part of #214 : the default font size is inherited from `body` and was indeed inconsistent with the size from `code` (and with the `line-height` attribute, I am not an artist so please tell me if you want to keep the `2px` difference between the two attributes).


yurug:
Do you have a screenshot of the new rendering? (To compare with the screenshot of #214.)

nobrakal:
Of course :
Screenshot_2019-05-22 demo - Learn OCaml v 0 10(3)

Issue #232, opened 2019-01-31
Uninformative internal error message must be improved


fpottier:
I am writing a new exercise. When I test it offline using `learn-ocaml grade`, it seems to work. When I try it inside the Web browser, grading fails with this impenetrable error message:

```
Erreur dans l'exercice lors du test de la solution utilisateur
```

This message *really needs to be improved* in order to explain what is going on.

Issue #226, opened 2018-12-27
Toplevel exception in student code is badly reported/explained


fpottier:
If the student raises an exception at the top level, e.g. by writing:
```
let x = failwith "todo"
```
then the grader reports a generic error,
```
Exercice échoué 0 pt
Erreur dans votre solution:
```
This seems inappropriate. Could learn-ocaml automatically produce a better report, indicating which exception is raised and where?

Incidentally, this would allow us to use `let x = failwith "todo"` in a template, to represent a hole that the student must fill. Perhaps we could even designate a special-purpose exception `ToBeImplemented "foo"` so as to indicate such a hole. The grader could then produce a better report, of the form `You have not yet implemented foo`.

PR #250, opened 2019-05-24
Print all buffers in case of error

nobrakal:
Hi,

While working on another issue, I hit the exact same than described in #232 : In the webapp, the grading fails with:
```
Erreur dans l'exercice lors du test de la solution utilisateur
```

And no other information.

So I decided to give it a try, and discovered that a buffer which contained some logs was not printed (and it was indeed the one where was the information I needed).

This PR fix this problem.

### DIFF
--- a/src/grader/grading.ml
+++ b/src/grader/grading.ml
@@ -84,10 +84,9 @@ let get_grade
           !flush_stderr () ;
           !flush_stdout () ;
           let msg =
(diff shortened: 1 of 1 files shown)


nobrakal:
Note, this also fix in part #226 , since the report also print now if an exception was raised by the student code.

yurug:
Fix #226 
Fix #232

PR #251, opened 2019-05-25
Global Download Button give a .tar archive of the repository

### LINKED ISSUE #162: Global "Download" button
Currently, in order to save information to the local computer, the options are:

- a "Download" button which is local to an exercise (and saves the source code as a .ml file)
- a global "Save" button which produces a dump of the whole session as a .json file

It would be nice to have a global "Download" button that produces the set of all .ml files for all exercises. Currently the way to achieve this is manually pressing "Download" on each exercise, which is tedious.

Asereme:
This resolves issue #162.
Actually, it doesn't catch errors of the command `git archive ...` for now.
And I have tried to make a .zip instead of .tar, but I have an encoding error which corrupts the archive.

Fixes #162


src/app/learnocaml_index_main.ml
```diff
@@ -848,6 +848,17 @@ let () =
...
   in
+  let download_all () =
+    let token = get_stored_token ()
+    and name = "archive.tar" in
+    Server_caller.request (Learnocaml_api.Archive_tar token) >|= function
+      | Ok tar ->
+          let contents = Js.string tar in
+          Learnocaml_common.fake_download ~name ~contents
+      | Error e ->
+        alert ~title:"DOWNLOAD ARCHIVE FAILED"
```
yurug: Please use "internationalized strings" for the title.

Asereme: ok, I forgot to do that

Asereme:
I have spotted the encoding issue: the "get" request from "lwt_request" will cast the response into a DOMString, which is utf-16 encoded, then the "Js.to_string" will cast into an Ocaml string by encoding it into utf-8, which could change its size.
I have tried to force casting into utf-16 with "Js.to_bytestring", but the "fake_download" function would need to create an object Url with a blob, which is utf-8 encoded

A workaround would be to get the response of the request as a blob directly and send it to the fake_download (WIP)

yurug:
Any news on this?

Asereme:
The workaround might need to create a "File.blob Js.t request" in the learnocaml_api if we want to avoid useless conversion. But I don't know if I should solve that way or if there are better solutions.

Asereme:
By creating a regular download URL (not the fake one with a blob), I don't have any issue with encoding and the web browser will detect the URL as a download by default.

PR #252, opened 2019-05-28
Add a simple authentication protocol when requiring a new token.

nobrakal:
Hi,

This PR change the API to add a simple authentication protocol when requiring a new token.

First of all, it adds a requirement for a config file `config_server.json` in the root of any repository.
This configuration file contains a secret (a `string`).
Then, when an user request a new token, it will be asked to provide the secret, which will be compared to the one on the server.

This feature allows to protect running instances from being used by non-authorized users.

I have some questions:

*  I don't know if making the config file mandatory is a good idea, certainly not since it breaks already existing repository, but on the other hand, not requiring it by default (ie. if there is no config file, build anyway and consider every new token request valid) will maybe lead users to ignore this security feature.
* If the config file exists but no secret are provided, any request for a new token is actually considered valid. I don't know if it is a desired behavior. It may be, since it provides a manual way to deactivate this security feature.
* Do we need to hash/encrypt anything ? I am not a security expert, and actually all secrets are transmitted and stored in clear.


yurug:
I think that the configuration should not be mandatory. If we have well-documented default values for every parameter, that is not a problem.

yurug:
Or: if the file does not exist, learnocaml could ask for a secret when it is launched and generate the configuration file automatically. An empty secret would mean "no secret".

yurug:
And yes : the secret must be hashed.

nobrakal:
@yurug  About the hash, which function do you want me to use ? I don't know (and can't find) many information about `Hashtbl.hash` so I am not sure. `Digest.string` does not seems viable for a cryptographic use since it is `MD5` under the hood.
There seems to be some implementations of `SHA` (like `ocaml-sha`), but not in pure OCaml, and `js_of_ocaml` is not very happy with it.

nobrakal:
@yurug I found a simple OCaml library made by OCamlPro some times ago (https://github.com/AltGr/ocp-sha/ ), I have "bundled" it in `src/utils/sha.ml`, and simplified it for our needs (not sure what to do with the License header of the file however).

Now:

* The `server_config.json` file is not mandatory anymore. If not present, it will be interpreted as "no secret", and anyone will be able to register, using any secret.
* If present, the secret (in clear in the repository) will be encrypted using SHA512 to be stored in the root of `www/`.

yurug:
You cannot use this implementation of SHA since it is released under GPL.

nobrakal:
Ah that is sad! Do you know which license I am able to bundle? I am not an expert in that field!

yurug:
Roughly speaking, almost anything that is free but not GPL is compatible with MIT.

nobrakal:
Ok, thanks for the clarification !

I originally bundled a library because there was no version of any library on opam that was compatible with versions of packages in learnocaml, but I just realized that I could require an older version of an opam package.

So I made learnocaml dependent on `digestif`, which provides a pure ocaml implementation (so compatible with js_of_ocaml) of sha512.

yurug:
Perfect then!

learn-ocaml.opam
```diff
@@ -1,6 +1,6 @@
 opam-version: "2.0"
 name: "learn-ocaml"
-version: "0.10"
+version: "0.11"
```
yurug: Why did you bump the version number?

src/repo/dune
```diff
@@ -4,7 +4,7 @@
  (modules Learnocaml_index
           Learnocaml_exercise)
  (libraries ocplib-json-typed
-            xor
+            learnocaml_xor
```
yurug: How is that change related to the PR?

yurug:
I am not sure to know if that is easy but could we check that the secret hash is only transmitted if HTTPS is activated? Otherwise, a "replay" attack seems easy.

yurug:
Another possibility would be to have the server send a "nonce" to be hashed with the hash of the secret.

learn-ocaml.opam
nobrakal: Because I introduced a change in the API, older clients won't be compatible with the new server (since I require a hash of the secret).

src/repo/dune
nobrakal: Digestif has itself a module named Xor, and dune complained about the presence of two modules with the same name.

yurug:
Thanks!

nobrakal:
Ah, you are right, I didn't thought about it, but the hash is part of the URL, so replay attacks are easy. I will make another PR to add a Nonce

PR #253, opened 2019-05-31
Better print for exceptions

### LINKED ISSUE #185: Printing exception values
OCaml's `Printexc.to_string`, which is used for printing exception values in test results, doesn't correctly print exceptions that have values associated with them. For example:

```
exception Error of int list ;;
exception Error of int list
Printexc.to_string (Error [1; 2; 3; 4]) ;;
- : string = "Error(_)"
```

This means the grader output for functions that are supposed to raise these kinds of exceptions is pretty useless. However, using the typed printer for exceptions does print them properly. i.e.:

```
let typed_printer ty ppf v =
  Introspection.print_value ppf v ty
let exn_to_string = Format.asprintf "%a" (typed_printer [%ty: exn])
```

Can the instances of `Printexc.to_string` in the code generating the reports be replaced by this instead?

nobrakal:
Hi,

This fix issue #185 , errors are now better rendered with functions from `Test_lib`.

I allowed myself to do some code factorization where I had to make some changes, please don't hesitate to tell me if you want me to revert these unrelated changes. 

Fixes #185

### DIFF
--- a/src/grader/test_lib.ml
+++ b/src/grader/test_lib.ml
@@ -745,17 +745,17 @@ module Make
             Learnocaml_report.[ Message ([ Text "Found " ;  Text k ; Text " " ; Code (pr n) ], Success 5) ]
           end
 
+      let print_exp = Format.asprintf "%a" Pprintast.expression
+      let stripper = ast_location_stripper.Ast_mapper.expr ast_location_stripper
+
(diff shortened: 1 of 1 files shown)


yurug:
Perfect. Thanks!

Issue #257, opened 2019-06-05
Error given by  learn-ocaml-client command line while grading when it doesn't compile is not very explicit


leunam217:
Dear learnocaml developers, 
I face an issue while attempting to use the learn-ocaml-client command line. I run a container running the server. Then I execute inside the container a grade command but I have an error. See atached files to see the errors, the repository and all the commands, 
Thank you in advance
[pfitaxel-server.zip](https://github.com/ocaml-sf/learn-ocaml/files/3256607/pfitaxel-server.zip)
![running the conatiner](https://user-images.githubusercontent.com/38218036/58948250-fad45700-8789-11e9-95bb-34a094e86bdc.png)
![executing a request](https://user-images.githubusercontent.com/38218036/58948259-fe67de00-8789-11e9-9f11-8b84864c144e.png)

erikmd:
@leunam217 by doing exactly the steps you describe, I can reproduce the issue.
However if I write a valid OCaml file by doing:

```
echo "let a = 1" >demo.ml
```

then it works:

<details>

```
~ $ learn-ocaml-client --token=X-OB9-G1S-OB4-GBZ --id=demo --server=http://localhost:8080 demo.ml
────────────────────────────────────────────────────────────────────────────────
## Exercise failed                                                   ❰   0 pt  ❱
────────────────────────────────────────────────────────────────────────────────


❰   0 pt  ❱ Function: plus
│  ❰   0 pt  ❱   Cannot find  plus
└─


❰   0 pt  ❱ Function: minus
│  ❰   0 pt  ❱   Cannot find  minus
└─


❰   0 pt  ❱ Function: times
│  ❰   0 pt  ❱   Cannot find  times
└─


❰   0 pt  ❱ Function: divide
│  ❰   0 pt  ❱   Cannot find  divide
└─


Results saved to server
```

</details>

So to sum up, the real issue here is that when the `demo.ml` file does not compile, the message should be more helpful than just saying:

```
[ERROR] Could not do the grading: Grading.User_code_error(_)
```

e.g. something like `ocamlc`'s output?

```
echo "test" >demo.ml
$ ocamlc demo.ml
File "demo.ml", line 1, characters 0-4:
Error: Unbound value test
```

nobrakal:
@yurug  It is strange that this was not automatically closed with the merge of my PR. Can you close it ?

yurug:
The syntax to claim that a PR closes an issue is a bit tricky.

PR #258, opened 2019-06-05
Better print of report errors in command-line client

nobrakal:
This fixes issue #257. It now prints effectively the compiler output when desired.

Thanks @erikmd for the details !

### DIFF
--- a/src/main/learnocaml_client.ml
+++ b/src/main/learnocaml_client.ml
@@ -650,9 +650,13 @@ let main o =
   if o.verbosity >= 1 then prerr_newline ();
   match report with
   | Error e ->
-      Printf.eprintf "[ERROR] Could not do the grading: %s"
-        (Printexc.to_string e);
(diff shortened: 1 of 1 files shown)


erikmd:
Thanks @nobrakal for your quick response!

@leunam217 could you test this new version on your example from #257 ?

By the way if you want to fetch the branch of this PR #258 before compiling it, you can use this git alias:

```bash
# once and for all
git config --global alias.pr '!f() { if [ $# -lt 1 ]; then echo "Usage: git pr <id> [<remote>]  # assuming <remote>[=origin] is on GitHub"; else git checkout -q "$(git rev-parse --verify HEAD)" && git fetch -fv "${2:-origin}" pull/"$1"/head:pr/"$1" && git checkout pr/"$1"; fi; }; f'

# then
cd .../learn-ocaml
git pr 258 origin
```

Issue #241, opened 2019-03-12
Some standard library modules are missing.


fpottier:
For some reason, several modules that exist in OCaml's standard library seem to be missing when executing student code and grading code. I have noted, for instance, that the following modules are missing: `Arg`, `Bigarray`, `Complex`, `Filename`, `Float`, `Lazy`, `Marshal`, `Queue`, `Scanf`, `Stack`, `Uchar`, `Weak`.

I can understand why some of these modules are missing; they might be difficult to implement in JavaScript in a browser. However, this should be documented. Furthermore, I cannot understand why several simple modules, which can be implemented entirely in OCaml, are missing: this includes `Queue`, `Stack`, and `Lazy`.

AltGr:
These could probably be added at learn-ocaml build time by just patching [src/grader/dune](https://github.com/ocaml-sf/learn-ocaml/blob/4e98d44c8e4647edef11560b630ca6699f55bc58/src/grader/dune#L72). As long as they are compatible with js_of_ocaml, of course.

fpottier:
Has there been any progress? I am still bothered by the lack of these modules, especially `Lazy`, `Queue`, and `Stack`.

adhameer:
Those three modules (and many of the ones you requested) are now available as of PR #260; a full list of the modules that were added is in the PR description. Of the ones you requested, I think the ones that are still missing are: `Arg`, `Bigarray`, and `Float`.

fpottier:
Thanks, I hadn't noticed! This is very useful.

PR #260, opened 2019-06-05
Add some missing modules from standard library

nobrakal:
Hi,

This fixes (in part) issue #241.
This PR makes available the following modules: `Complex`, `Filename`, `Lazy`, `Marshal`, `Queue`, `Scanf`, `Stack`, `Uchar` and `Weak`.

Adding `Float` and `Bigarray` seems possible (see https://ocsigen.org/js_of_ocaml/3.1.0/manual/overview for bigarray), but it goes beyond my copying/pasting skill.

I allowed myself to order the list of modules in alphabetical order, to simplify the reading.

### DIFF
--- a/src/grader/dune
+++ b/src/grader/dune
@@ -47,29 +47,28 @@
 
 (rule
  (targets embedded_cmis.ml)
- (deps %{ocaml-config:standard_library}/pervasives.cmi
-       %{ocaml-config:standard_library}/array.cmi
+ (deps %{ocaml-config:standard_library}/array.cmi
        %{ocaml-config:standard_library}/arrayLabels.cmi
-       %{ocaml-config:standard_library}/list.cmi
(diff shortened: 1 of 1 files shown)


yurug:
Thanks!

Just a remark: in general in OCaml, modules can have side effects and for this reason, linking order is important. Reordering can have a semantic impact.

Here, that is fine because the modules are side-effect free.

lsylvestre:
Hi, 

While we are at it, do you mind adding CamlinternalOO and CamlinternalMod ?

```
%{ocaml-config:standard_library}/camlinternalOO.cmi
%{ocaml-config:standard_library}/camlinternalMod.cmi
``` 
(https://github.com/ocaml-sf/learn-ocaml/blob/master/src/grader/dune#L56)

Thanks!

Issue #202, opened 2018-12-02
Placement of toplevel clear/reset buttons


adhameer:
It is really easy to click the clear/reset buttons in the toplevel by accident while trying to click into the text box to type some code. Maybe those buttons could be moved to the top-right corner or something... they aren't really buttons you would be clicking extremely often.

nobrakal:
@yurug I think this issue can be closed

adhameer:
I think it would be nice to have the Eval Phrase button on the right of the text box instead of in the top right with everything else. That would make it more clear what it actually does. It's not very intuitive to have it so far away from the text field.

EmileRolley:
Closed by #261

PR #261, opened 2019-06-06
Placement toplevel buttons

beobillly:
issue #202

I propose this solution.
It works on different size and with mobile format. Buttons are placed on top of the toplevel pane so they are always visible.

the toplevel input has been moved at the bottom of the window

toplevelplacement

toplevelplacement2


yurug:
There is a risk that the buttons hide the text, right?

beobillly:
It's automatically resized depending on the size of the screen.
The buttons can't hide the text. At worst they are just above it.
If you found it's too close I can arrange that.

Also I just found out that I did that for the toplevel of the Exercise menu but not for the toplevel of Activities menu, would you like me to do the same for it ?

toplevel of Activities
toplevel of activities

toplevel of Exercises
toplevel of exercise

beobillly:
I changed the placement of the Activities toplevel buttons at the same place than the Exercices.

yurug:
Thanks!

PR #266, opened 2019-06-12
feat (print_server) : Add print_server option to the command line 

leunam217:


leunam217:
It will be incompatible with #267 so better i remove this and start with #267 as base

Issue #264, opened 2019-06-11
Add print-server option to learn-ocaml-client


leunam217:
If a user is making exercises from different servers it's good if he can know what server is he submitting the exercise.

leunam217:
I'm going to prepare a PR for this issue…

leunam217:
Can someone merge PR #269 to fix this issue please?

PR #267, opened 2019-06-12
Refactoring client and adding a create-token command

### LINKED ISSUE #262: Adding --create-token option to learn-ocaml-client
Hi , I'm implementing an editor plugin and i have to manage token creation. It would be easier for me if i could do 
learn-ocaml-client --set-options --create-token=test --server=http://test.com
to create a token associated with the nickname test.

nobrakal:
Hi,

This PR introduces big changes on the syntax of `learn-ocaml-client`. To say it quickly, it adds commands to the client, and particularly one to create tokens (so this PR fixes #262).

The client seems to have been built to do a single task: grading file on a server. Now, it seems to have evolved, and there are many possible contradictory options (for example, a valid invocation is `learn-ocaml-client --fetch --console`, but has no sens, since the program will just fetch solutions and exit).

So, using `Cmdliner`, I restricted the use of options and added commands (the syntax looks like the one from `git`) to organize things. With this PR, the MAN page looks like:
```
NAME
       learn-ocaml-client - Learn-ocaml grading client

SYNOPSIS
       learn-ocaml-client COMMAND ...

DESCRIPTION
       Learn-ocaml-client, default command is grade

COMMANDS
       create-token
           Create a token

       fetch
           Fetch the user's solutions

       grade
           Learn-ocaml grading client

       print-token
           Just print the configured user token

       set-options
           Set configuration
```

So you can specify which command to run, and after use dedicated options if needed. For example, here is the output of `learn-ocaml-client create-token --help`:

```
NAME
       learn-ocaml-client-create-token - Create a token

SYNOPSIS
       learn-ocaml-client create-token [OPTION]... [NICKNAME] [SECRET]

DESCRIPTION
       Create a token on the server with the desired nickname

ARGUMENTS
       NICKNAME
           The desired nickname

       SECRET
           The secret. If not provided, use "" as a secret

OPTIONS
       --help[=FMT] (default=auto)
           Show this help in format FMT. The value FMT must be one of `auto',
           `pager', `groff' or `plain'. With `auto', the format is `pager` or
           `plain' whenever the TERM env var is `dumb' or undefined.

       -s URL, --server=URL (absent LEARNOCAML_SERVER env)
           The URL of the learn-ocaml server

       --version
           Show version information.
```

To keep things compatible, calling `learn-ocaml-client` without any command has the effect of calling `learn-ocaml-client grade` (for example, I did not have to change the test-suite, which use the client to grade files).

Note that, except for `create-token`, this PR does not add capabilities to the client. This change open the path to adding more capabilities to the client, since any command can now be added without breaking the others.

I tried to organize the `learnocaml_client.ml` file with modules, one for each command and for each arguments set. Anyway, I wonder if we should split this file: having 800 lines makes it difficult to read, even with modules to organize tings.

### DIFF
--- a/src/main/dune
+++ b/src/main/dune
@@ -34,6 +34,7 @@
             cohttp.lwt
             grading_cli
             learnocaml_data
+	    learnocaml_store
             learnocaml_api)
 )
 
--- a/src/main/learnocaml_client.ml
+++ b/src/main/learnocaml_client.ml
@@ -56,7 +56,7 @@ module Args_global = struct
 
   let local =
     value & flag & info ["local"] ~doc:
-      "Generate a configuration file local to the current directory, rather \
+      "Use a configuration file local to the current directory, rather \
        than user-wide"
 
   let apply server_url token local =
@@ -78,7 +78,7 @@ module Args_create_token = struct
 
   let secret =
     value & pos 1 string "" & info [] ~docv:"SECRET" ~doc:
-      "The secret"
+      "The secret. If not provided, use \"\" as a secret"
 
   let apply nickname secret = {nickname; secret}
 
@@ -669,7 +669,7 @@ module Grade = struct
     Term.(
       const (fun go eo -> Pervasives.exit (Lwt_main.run (grade go eo)))
       $ Args_global.term $ Args_exercises.term),
-    Term.info ~version ~man
+    Term.info ~man
       ~doc:"Learn-ocaml grading client"
       "grade"
 end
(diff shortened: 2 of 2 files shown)


nobrakal:
Ah, for example, a command can be added to address issue #264

leunam217:
So I will add the print-server command for #264 and create a pool request based on this branch

yurug:
The patch is a bit long to review but it seems OK. Thanks!

PR #269, opened 2019-06-12
 Add print_server command to the client 

leunam217:


src/main/learnocaml_client.ml
```diff
@@ -698,6 +698,23 @@ module Print_token = struct
       "print-token"
 end
 
+module Print_server = struct
+  let print_server o =
+    get_config_o o
+    >>= fun config ->
+    Lwt_io.printl (Uri.to_string config.ConfigFile.server)
+    >|= fun () -> 0
```
yurug: Can you explain this `0`?

nobrakal: I think I can answer: it is the return code of the application

leunam217: I think that it's the return value but i have to admit that i didn't search for it. I just did the same that print_token does.
https://github.com/ocaml-sf/learn-ocaml/blob/d036297cffaa82f4c1f783174268d8b087f6728e/src/main/learnocaml_client.ml#L690

PR #270, opened 2019-06-12
Add a nonce to remove vulnerability to replay attack

nobrakal:
Hi,

As suggested by @yurug, this PR add a nonce mechanism in addition of the secret when requiring a new token. The protocol is now:

* Ask a nonce to the server (which is a random string of 20 char)
* Compute a hash of what you think is the secret and the nonce
* Send to the server the request

Please note that a nonce is valid only for one request, and is linked to the IP address of the client (I don't know if this is a limitation, but I don't know either how to identify a client without a token other than with his ip).

With this protocol, replay attack should be difficult.

Also note that the whole process is hidden to the end-user.

### DIFF
--- a/src/server/learnocaml_server.ml
+++ b/src/server/learnocaml_server.ml
@@ -179,14 +179,30 @@ module Request_handler = struct
   let alphanum = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789"
   let alphanum_len = String.length alphanum
 
-  let nonce_req : (Ipaddr.t option, string) Hashtbl.t = Hashtbl.create 533
+  let nonce_req : (string, string) Hashtbl.t = Hashtbl.create 533
 
   let token_save_mutex = Lwt_utils.gen_mutex_table ()
 
-  let ipaddr_of_endp =
+  let rec string_of_endp =
     function
-    | `TCP (i,_) -> Some i
-    | _ -> None
+    | `TCP (i,_) -> Some (Ipaddr.to_string i)
+    | `Unix_domain_socket s -> Some s
+    | `Vchan_direct (i,_) -> Some (string_of_int i)
+    | `Vchan_domain_socket (u,v) -> Some (u ^ v)
+    | `TLS (s,e) ->
+       begin
+         match string_of_endp e with
+         | Some s' -> Some (s ^ s')
+         | None    -> None
+       end
+    | `Unknown _ -> None
+
+  let valid_string_of_endp e f =
+    match string_of_endp e with
+    | None ->
+       Lwt.return
+         (Status {code = `Forbidden; body = "No address information avaible"})
+    | Some s -> f s
 
   let callback_raw: type resp. Conduit.endp -> string option -> caching -> resp Api.request -> resp ret
     = fun conn secret cache -> function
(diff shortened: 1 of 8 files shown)


src/server/learnocaml_server.ml
```diff
@@ -176,27 +176,55 @@ module Request_handler = struct
     | Response ({contents; _} as r) -> Response {r with contents = f contents}
     | (Cached _ | Status _) as r -> r
 
+  let alphanum = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789"
+  let alphanum_len = String.length alphanum
+
+  let nonce_req : (Ipaddr.t option, string) Hashtbl.t = Hashtbl.create 533
+
```
yurug: Hmm, why do you need an `option` in the type of key?

src/server/learnocaml_server.ml
```diff
@@ -176,27 +176,55 @@ module Request_handler = struct
...
-                                  body = "Bad secret"})
-         else
-           Token.create_student () >>= fun tok ->
-           (match nick with None -> Lwt.return_unit | Some nickname ->
-             Save.set tok Save.{empty with nickname}) >>= fun () ->
-             respond_json cache tok
+         let conn = ipaddr_of_endp conn in
+         begin
+           let forbid s = Status {code = `Forbidden; body = s} in
+           match Hashtbl.find_opt nonce_req conn with
```
yurug: This code makes me wonder if you do not need an error monad here to improve readability.

src/state/dune
```diff
@@ -20,6 +20,7 @@
  (modules Learnocaml_version Learnocaml_api)
  (libraries ocplib-json-typed
             ezjsonm
+	    conduit
```
yurug: What is `conduit`?

src/server/learnocaml_server.ml
nobrakal: I am not sure we always have access to an IP, (I use the result of `ipaddr_of_endp`, just below, and I am not sure what to do if I don't have a TCP connection. The other possibilities are here: https://github.com/mirage/ocaml-conduit/blob/master/lib/conduit.mli#L52 ).

src/state/dune
nobrakal: It is the library used by `cohttp` to abstract connections.

src/server/learnocaml_server.ml
yurug: Couldn't you use the `endp` as a key? (After filtering the `Unknown case, of course as it should result in a user error.)

nobrakal: The problem is that `endp` contains also the port for TCP, and the same client can send request from the same ip, but its output port is not guaranteed to be same (I hit this issue, and I switched to use only the ip).

src/server/learnocaml_server.ml
nobrakal: I agree, but how do you use an error monad in Ocaml ? This refer to a pretty particular thing in Haskell, and I am not sure how to do it in Ocaml.

src/server/learnocaml_server.ml
yurug: I see. In any case, the `None` key can be reused several times by two clients with no IP. That's a security problem, probably.

Couldn't you write a function of type `endp -> string` that would improve this situation (string being used as type for keys)?

src/server/learnocaml_server.ml
yurug: Nothing really special:
https://gist.github.com/yurug/0d0ab0682a11a91a0648b47c036df8ea

By the way, with ocaml 4.08, we will be able to say: 
``
let (let*) x f = bind x f
``
so that `let* x = t in u` will be equivalent to `bind t (fun x -> u)`.

src/server/learnocaml_server.ml
nobrakal: You are right, I have made a `string_of_endp`

src/server/learnocaml_server.ml
nobrakal: Ah thanks for the example, I see (nice new syntax !). Just to be sure, do you want me to change the whole function (`callback_raw`) to be encapsulated in an error monad (maybe something with `either` in place of `option`, since we want to keep the string and the code of the status when failing) ?

yurug: Yes. Every change that could improve the readability of this code is welcome.

nobrakal: Ok, since I changed a bit the `response` type to extract to error case and the `callback_raw` function which gives now a result of type `(resp response, error) result Lwt.t` (with `error = (Cohttp.Code.status_code * string)`.

I defined a useful bind : `( >?= ) : 'a result Lwt.t -> ('a -> 'b result Lwt.t) -> 'b result Lwt.t`, with a strange name to avoid name collision with`Lwt.>>=`.

The whole thing seems more readable to me (even if a do-notation would be welcome :) )

PR #274, opened 2019-06-13
Add other modules

nobrakal:
Hi,

This answer a request by @lsylvestre (made in #260 ).

It makes available to user code two more modules. `json_of_ocaml` does not seems to have problems with them.


yurug:
Which issue are you referring to?

nobrakal:
There is no issue, only a comment made in https://github.com/ocaml-sf/learn-ocaml/pull/260#issuecomment-500826290 .

PR #276, opened 2019-06-14
Fetch exercises list

### LINKED ISSUE #275: Re-add option --id to learn-ocaml-client-fetch
In the current version of learn-ocaml-client there is no --id option anymore for the fetch command. I think this option is useful if the user have 5 exercises and wants just to download 1 file.

nobrakal:
Hi,

This fixes #275. It adds the capability to the client to `fetch` only some saves. The syntax looks like:

```
SYNOPSIS
       learn-ocaml-client fetch [OPTION]... [EXERCISE_ID]...

DESCRIPTION
       Fetch the user's solutions on the server to the current directory

ARGUMENTS
       EXERCISE_ID
           Exercise identifier. Can be repeated. If not present, all the
           exercises will be downloaded
```

### DIFF
--- a/src/main/learnocaml_client.ml
+++ b/src/main/learnocaml_client.ml
@@ -764,7 +764,9 @@ module Fetch = struct
     | [] -> Lwt.return_unit
     | xs ->
        Lwt_list.iter_s
-         (fun id -> Lwt_io.printl ("Exercise " ^ id ^ " was not found on the server"))
+         (fun id ->
+           Lwt_io.printl
+             ("Warning: exercise " ^ id ^ " was not found on the server"))
          xs
 
   let fetch o lst =
--- a/src/main/learnocaml_client.ml
+++ b/src/main/learnocaml_client.ml
@@ -154,6 +154,15 @@ module Args_exercises = struct
           $color_when $verbose )
 end
 
+module Args_fetch = struct
+  let id =
+    value & pos_all string [] & info [] ~docv:"EXERCISE_ID" ~doc:
+      "Exercise identifier. Can be repeated. \
+       If not present, all the exercises will be downloaded"
+
+  let term = Term.(const (fun x -> x) $ id)
+end
+
 module ConfigFile = struct
 
   type t = {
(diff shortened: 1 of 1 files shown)


src/main/learnocaml_client.ml
```diff
@@ -154,6 +154,15 @@ module Args_exercises = struct
           $color_when $verbose )
 end
 
+module Args_fetch = struct
+  let id =
+    value & pos_all string [] & info [] ~docv:"EXERCISE_ID" ~doc:
+      "Exercise identifier. Can be repeated. \
+       If not present, all the exercises will be downloaded"
```
yurug: Please add a final dot.

src/main/learnocaml_client.ml
```diff
@@ -731,31 +740,50 @@ module Fetch = struct
            (Token.to_string token)
       | e -> Lwt.fail e
 
-  let write_save_files save =
-  Lwt_list.iter_s (fun (id, st) ->
+  let write_save_files lst save =
+    let pred x = match lst with
```
yurug: This `pred` function seems weird. Can you describe its purpose?

src/main/learnocaml_client.ml
nobrakal: Speaking of it, there is no consensus in the code, some of the descriptions have a final dot (`--version`, `--help`), but not the others (`--server`, `--token`, for example).

src/main/learnocaml_client.ml
nobrakal: I am not good at choosing names :) It is a predicate to decide if we have to get the file or not. If no files where specified (hence an empty `lst`), we have to get all files, so the predicate is always true.

nobrakal:
Oh, while I am at it, I will fix #278 , please don't merge this for now.

src/main/learnocaml_client.ml
yurug: I prefer a final dot everywhere then.

src/main/learnocaml_client.ml
yurug: Then, we need a comment to explain this.

yurug:
Are you done with this PR, @nobrakal ?

nobrakal:
Oops, yes indeed, all is good for me !

Issue #140, opened 2018-09-15
Prompt to refresh when server restarted


adhameer:
When leaving an exercise open in a tab while the server is restarted, grading an exercise will use the old version of the grader and sync those results to the server with no notification to the student that the grader may have been updated. This is especially critical if we have restarted the server because we added additional tests to the grader.

It would be nice to have a pop-up when running the grader in this case saying something like, "The server has been updated, please refresh the page to make sure you are using the latest version of Learn OCaml (none of your work will be lost)."

It may be impossible to do something specifically when running the grader because I think that is all local, but running the grader calls sync as well, which makes a request to the server - maybe the information needed to check if the server has been restarted can be passed in the response?

Asereme:
It would be possible to generate randomly a server ID during the building to check if it has been updated.
And a new request "ServerId" can be created in the API to get the current ID.

EmileRolley:
Closed by #286

PR #286, opened 2019-06-25
Refresh update

Asereme:
This merge solve the issue #140.
Every "learn-ocaml build" will generate a random integer as "server_id" in the "server_config.json" of the app directory.
Each time a user will access to an exercise, a local "server_id" will be registered in the local storage.
And at every grade, a request will be sent to get the "server_id" in the server and compare with the local one.


yurug:
Seems nicely done! Can you give us a screenshot of the current user dialog that bumps to alert the user that a reload is needed?

Asereme:
out
Notice that the alert do not suspend the loading of the grader, and refresh the page can prevent the grade from synchronizing.

PR #284, opened 2019-06-24
Playground activity

nobrakal:
Hi,

This (big) PR fixes #277.

Mainly, it adds a playground activity, with, as requested, only the editor and a top-level. A playground is only specified by a `template.ml` file (and maybe a `prelude.ml`, made in prevision of #254 ), and one can attach a tittle and a short description if desired, to be displayed in the selection panel.

A `playground` repository is made after an `exercises` one, and has a structure like:

```
playground
├── index.json
├── monad_list
│   ├── prelude.ml
│   └── template.ml
└── monad_option
    └── template.ml
```

and is rendered as:

Screenshot_2019-06-24 Learn OCaml v 0 11(3)

Screenshot_2019-06-24 monad_option - Learn OCaml v 0 11

Note that actually, user code is never sent to the server (there is no "sync" button), and no token is ever required.

Since all of this is based on how exercises are processed, I tried to factorize code where I was able to.

I added an example in `demo-repository`, but I don't know where to document the change...

PR #287, opened 2019-06-27
Partitioning codes implementing the same function

nobrakal:
Hi,

Pretty big PR here.

It adds a method (available through the API) to classify codes implementing the same function. The goal is to help teachers to quickly view how an exercise was solved by its students, without going through hundred of copies.

The tool focuses on classifying codes _of the same function_.

Roughly, this is achieved by:

* Splitting codes by their grade (if there is any)
* Then, for each obtained class, try to identify "close" codes. This is done by reasoning on their AST, and make some clustering. I explained the whole process in a [blog post](https://blog.nyarlathotep.one/2019/06/learnocaml-code-classification/).

The result is the a data:
```OCaml
  type t =
  {
    not_graded : Token.t list;
    bad_type   : Token.t list;
    patition_by_grade :
      (int *
         (((Token.t * string) list) tree list))
        list;
  }
```

Containing the codes that were not graded, those with a type different than the one from the function used in `solution.ml`, then a partition where for each grade, there is a list of classes, and for each class, a tree (from the clustering) of tokens and their implementation (only of the concerned function), representing the class.

The tree represents "equivalent" (as the tool detects it) codes and his leaves are constituted of tokens that share "exactly the same" code (up to inlining, variable and function name).

I am willing to share screenshots of the rendered result, but the only database I have contains tokens that are still valid on http://ocaml.hackojo.org, and some students names, so I don't think it is possible.

The important thing is that, on the teacher tab, one can use the middle mouse button on an exercise name, enter the function name, and the result will be displayed on an other page.

@yurug I am hitting a strange problem: the OCaml typechecker is printing information on stdout (about missing cases in pattern-matching for example), and I don't understand how to disable this output (because it floods the logs of the server). Is there a function that "mute" totally the standard output of the program ?

There are some codes that can be factorized, but I will wait the merge of #284 (where I already did a lot of factorization) for it.

I tried to comment as much as possible my code, but don't hesitate to ask clarifications!


nobrakal:
Hi @yurug , I have unbundled the partitioning code that now live in is own library, [asak](https://opam.ocaml.org/packages/asak/).

As said, I have factorized more code.

I don't really now how to explain more how to use the tool, so please don't hesitate to ask.

PR #290, opened 2019-07-01
Add a facility for grading using a postcondition

### LINKED ISSUE #223: Add a facility for testing a function against a postcondition
The module `Test_lib` offers `test_function_1`, which compares an actual output against an explicit
expected output, and `test_function_1_against`, which compares an actual output
against the output of a reference implementation. We seem to be lacking
`test_function_1_post`, which would test whether an actual output satisfies a
user-specified postcondition.

Such a facility seems necessary when the specification is nondeterministic (i.e., when multiple outputs are correct).

At the moment, I have manually emulated this function using lower-level functions (`test_value`, `lookup_student`, `run_timeout`, etc.), but it would be preferable if the library offered this facility.

nobrakal:
Hi,

This fixes #223 .

It adds `test_function_X_against_postcond` functions, which are taking in particular an argument of type `'a Ty.ty -> 'a result -> Learnocaml_report.t`, that is, a function that is capable to produce a report from the student code output.

Maybe something of type `'a Ty.ty -> 'a result -> bool` will be more user friendly, but it is difficult to control how many point you give.

### DIFF
--- a/src/grader/test_lib.mli
+++ b/src/grader/test_lib.mli
@@ -765,7 +765,18 @@ module type S = sig
                 -> ('c * string * string)
                 -> Learnocaml_report.t) ->
       ?sampler : (unit -> 'a * 'b) ->
-      ('a -> 'b -> 'c) Ty.ty -> string -> ('a * 'b) list -> Learnocaml_report.t
+                 ('a -> 'b -> 'c) Ty.ty -> string -> ('a * 'b) list -> Learnocaml_report.t
+
+    (* TODO *)
+    val test_function_2_against_postcond :
+      ?gen: int ->
+      ?test_stdout: io_postcond ->
+      ?test_stderr: io_postcond ->
+      ?before_reference : ('a -> 'b -> unit) ->
+      ?before_user : ('a -> 'b -> unit) ->
+      ?after : ('a -> 'b -> ('c * string * string) -> Learnocaml_report.t) ->
+      ?sampler : (unit -> 'a * 'b) ->
+      'c postcond -> ('a -> 'b -> 'c) Ty.ty -> string -> ('a * 'b) list -> Learnocaml_report.t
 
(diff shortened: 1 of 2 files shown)


nobrakal:
I addressed the comment of @adhameer on #290 : the postcondition has now access to the parameters of the test.

I have just a side question. The postcondition's type for unary function is:

```OCaml
'a -> 'b Ty.ty -> 'b result -> Learnocaml_report.t
```

The `'b Ty.ty` was taken by a copy/paste from the `'a tester` type:

```OCaml
type 'a tester =
    'a Ty.ty -> 'a result -> 'a result -> Learnocaml_report.t
```

I don't understand what is the use of the `'a Ty.ty` parameter. Why the postcondition (or the tester) should be given this ?

src/grader/test_lib.mli
```diff
@@ -664,6 +669,22 @@ module type S = sig
       ?sampler : (unit -> 'a) ->
       ('a -> 'b) Ty.ty -> string -> 'a list -> Learnocaml_report.t
 
+    (** [test_function_1_against_postcond postcond ty name tests] tests that
+       the function named [name] statisfies the postcondition [postcond].
```
yurug: statisfies -> satisfies

src/grader/test_lib.mli
```diff
@@ -835,6 +872,22 @@ module type S = sig
       -> string -> ('a * 'b * 'c) list
       -> Learnocaml_report.t
 
+    (** [test_function_3_against_postcond postcond ty name tests] tests that
+       the function named [name] statisfies the postcondition [postcond].
```
yurug:
```suggestion
       the function named [name] satisfies the postcondition [postcond].
```

src/grader/test_lib.mli
nobrakal: Oops

Issue #207, opened 2018-12-12
Grading: giving negative points


adhameer:
Sometimes we want to take off points from a student for making some mistake, but don't necessarily want to give the student points for _not_ making that mistake. This can currently be done by creating a `Success` report with a negative value, but the background of such a report is still green, which is confusing.

Checking whether the points value for a `Success` report is negative and changing the background colour accordingly could work, although the name "Success" would no longer make sense; and adding a points value to `Failure` reports would break a lot of existing code. Perhaps a new kind of status ("Penalty"?) could be added?

yurug:
This is a great idea. Thank you @adhameer.

Asereme:
Should the `Penalty` status be considered as a bad `Success` or a `Failure` ?
Or should we consider the function with penalty as `Complete` or `Imcomplete` ?

adhameer:
My preference is for it to be considered as Failure, and the function with a penalty should be considered Incomplete.

Asereme:
I have something like this.
![out](https://user-images.githubusercontent.com/50735325/60603977-09c51e00-9db7-11e9-932d-d3fb03970f0a.png)
Also, should we allow negatives points in a function or set a minimum of 0?
By default, an exercise has a minimum points of 0.
![out](https://user-images.githubusercontent.com/50735325/60605136-3da14300-9db9-11e9-8d4d-ec4a1b245bf8.png)

adhameer:
Hmm. Is it possible to make the minimum score a per-section property? I can think of both cases where I would want the minimum score for a section to be 0 (such as in the example you have given) and where I would want the overall score for a section to be negative (for example, if I want to grade code style, I would probably have separate section called "Code Style" which might contain a `Penalty`, and I would like that grade deduction to be applied to the overall score for the exercise).

It would also be nice to have the point deduction indicated in another colour, maybe red to match the failures.

erikmd:
I +1 this feature request.
In our assignments we have often used negative scores `n<0` associated to `Success n`, especially to implement a separate section related to "Code quality". E.g. if a student writes `if b then true else false` one or more times in the code (which is experimentally a very frequent pattern), he'll gets `-2` points or so, globally on the note.

So it'll be essential for us to keep the ability to have a section with a negative score [i.e., not imposing `max(0,⋅)`]

Asereme:
To avoid breaking tests, I will create a new constructor `SectionMin of text * t * int` which will work as a Section with a minimum score.
[...]

PR #292, opened 2019-07-04
Negative points & Section with minimum points

Asereme:
This PR fix issue #207 by adding formal negative points with `Penalty` constructor and a special section `SectionMin of text * t * int` which is a `Section` with a minimal score (negative or positive).


Asereme:
out

Seems like Travis worked for me but not here. Maybe we should relaunch Travis.

yurug:
Thanks!

PR #298, opened 2019-07-12
Some convenience functions for graders

adhameer:
This PR includes some functions we often copy and paste into all our graders and which we thought would be useful to incorporate into the grader library. (Please feel free to suggest better names for the conditional report generation operators `@@>` and `@@=`.)


yurug:
Thanks!

PR #304, opened 2019-08-11
Fix CI build

erikmd:
Dear @yurug, it seems current master [didn't build anymore](https://travis-ci.org/ocaml-sf/learn-ocaml) because combining #292 and #287 led to some non exhaustive pattern-matching.

This small PR solves this (please review whether my fix is reasonable). It also updates the Travis CI markdown badge to make it "clickable".


yurug:
Perfect! Thank you @erikmd!

PR #306, opened 2019-08-11
Mutation testing functionality

adhameer:
This PR adds mutation testing functionality for requiring students to write test suites for their functions, and automatically grading the quality of these test suites based on desired test scenarios provided by the instructor.

When this extension is enabled, students can write test cases (input/expected output pairs) for a function `foo: 'a -> 'b` in a variable called `foo_tests: ('a * 'b) list`. These tests are first run against the solution code to ensure that they are correct tests. They can optionally also be run against the student's own submission. Finally, the tests are run against a selection of buggy programs specified by the instructor. The quality of the student's tests is judged by how many of the buggy programs fail at least one of the student's tests. (The ideal is for each buggy program to fail at least one of the student's tests, meaning that the student's test suite is rigorous enough to catch each of the given bugs.)

The mutation test functions are designed to look similar to the correctness testing functions in `Test_lib`; a typical invocation looks like this:

```
Mutation_test.test_unit_tests_2
    [%ty: int -> int -> int] "plus"
    (* List of buggy implementations *)
    [("Subtracts instead of adding", fun x y -> x - y)]
```

The demo exercise has been updated to demonstrate this functionality.

I suppose it is probably preferable to add this to the `Test_lib` file instead of having it in its own file; let me know.

### DIFF
--- a/src/grader/mutation_test.mli
+++ b/src/grader/mutation_test.mli
@@ -26,6 +26,7 @@ val run_test_against_mutant: ('a -> 'b) -> ('a * 'b) -> bool
 (** Running mutation tests on a student's test suite.
     For testing a function called [foo], the student's tests should
     be in a variable called [foo_tests].
+
     This module needs to be instantiated with an instance of
     Test_lib, which is available to the grader code:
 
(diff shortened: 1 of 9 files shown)


src/grader/mutation_test.mli
```diff
@@ -0,0 +1,110 @@
+(* This file is part of Learn-OCaml.
+ *
+ * Copyright (C) 2019 OCaml Software Foundation.
+ * Copyright (C) 2016-2018 OCamlPro.
+ *
+ * Learn-OCaml is distributed under the terms of the MIT license. See the
+ * included LICENSE file for details. *)
+
```
yurug: Please add a documentation header to present the purpose of this module.

src/grader/mutation_test.mli
```diff
@@ -0,0 +1,110 @@
...
+type 'a test_result =
```
yurug: This looks like a quite general type definition. Shouldn't it be in another module?

src/grader/mutation_test.mli
```diff
@@ -0,0 +1,110 @@
...
+  | Pass
+  | Fail of 'a
+  | Err of exn
+
+(** A mutant is a pair of a string describing the mutant and the
+    mutant function itself.
```
yurug: Please start with a definition of what a mutant function is, and the problem it solves.

src/grader/mutation_test.mli
```diff
@@ -0,0 +1,110 @@
...
+(** Run a test (a pair of input and expected output) on a function.
+    The [compare] parameter specifies a comparison function for
+    comparing the expected and actual outputs, and defaults to
+    structural equality ([(=)]).
+*)
+val run_test_against:
+  ?compare: ('b -> 'b -> bool) ->
+  ('a -> 'b) -> ('a * 'b) -> 'b test_result
+
+(** Run a test (a pair of input and expected output) on a mutant.
```
yurug: Given this first sentence, it is a bit weird not to see a `mutant` as an argument of this function.

src/grader/mutation_test.mli
```diff
@@ -0,0 +1,110 @@
...
+    should be given for each mutant exposed by the test suite.
+    If [test_student_soln] is [true] (as it is by default),
+    also runs the student's test suite against the student's own
+    implementation and reports the results.
+    The [test] parameter specifies a comparison function for the
+    expected and actual outputs, and defaults to structural
+    equality ([(=)]).
+*)
+module type S = sig
+  val test_unit_tests_1:
```
yurug: You could profit from #302 to avoid the multiple versions of this function.

src/grader/mutation_test.mli
adhameer: Hmm, it is probably not necessary to expose this type and the associated `run_test_against` functions from this module at all, since it is basically just a weaker version of the test functions in `Test_lib`.

adhameer:
Updated, addressing comments from the review and a couple other points:
- Allow instructors to specify scores on a per-mutant basis instead of weighting them all equally
- Rename the type `mutant` to `mutant_info` which is more accurate
- Respect the timeout value by enclosing calls to student code in `run_timeout` instead of calling the functions directly
- Add documentation and reorganize the module signature a little

src/grader/mutation_test.mli
adhameer: The module signature was designed around the test functions we used most often, which are the only ones documented in the tutorials (`test_function_1_against_solution`, etc.). I'm not familiar with the version that uses function prototypes so I am not sure I can write a version like `test_unit_tests` using function prototypes at this moment...

I think regardless it is useful to keep the multiple versions to match the `Test_lib` functions, right?

yurug:
Thanks a lot for this contribution!

PR #299, opened 2019-07-12
feat : add more flexibility to the client cli

leunam217:
The print-server and print-token commands don't ask you to enter information if there is not a config file.

PR #310, opened 2019-09-09
[minor] Fix warning 29

gasche:
This fixes a minor warning message in `src/main/learnocaml_client.ml` that was introduced in #299.

I also changed the formulation of the documentation from

> You must provide a token or a nickname and a secret

(is it "(a token or a nickname) and a secret"?) to 

> You must provide a token or a nickname+secret pair


gasche:
Gentle ping: this is a very small change and it would be nice to not have the warning anymore when building from master.

yurug:
Thanks!

PR #318, opened 2019-10-02
introspection feature: sampling pairs

gasche:
On the current master branch, writing a test such as

```ocaml
test_function_1_against_solution ~gen:3
  [%ty: bool * int -> int * bool] "swap" []
```

results in a build-time failure:

```
Exception: Failure "unsamplable type".
Called from file "src/grader/introspection.ml", line 220, characters 30-41
Called from file "src/grader/introspection.ml", line 220, characters 10-41
Called from file "src/grader/test_lib.ml", line 1291, characters 27-54
[...]
```

This PR fixes this problem by teaching the grader how to generate
random values of pair type.

Note that in only works for pairs (2-ary tuples), not tuples of
arbitrary arity.

### DIFF
--- a/src/grader/introspection.ml
+++ b/src/grader/introspection.ml
@@ -205,13 +205,21 @@ let sample_value ty =
     let open Asttypes in
     let open Types in
     let open Ast_helper in
+    let sampler_id suffix =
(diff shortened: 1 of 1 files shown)


yurug:
Thanks for the feature (and the code factorization)!

PR #324, opened 2019-10-06
Learnocaml_store: refactor {get_from,write_to}_file

gasche:
The same operations of file+json (de)serialization were used many times in the Learnocaml_store module.

This PR refactors it into generic operations used within the module,
and also by learnocaml_main as a client (to (de)serialize configuration values).

PR #325, opened 2019-10-06
Refactor Learnocaml_data.Server

gasche:
This PR sits on top of #324, so it would be easier to review if #324 was merged.

---

The previous API (from #286) was unclear; I couldn't understand the role of
enc_init without reading the client code, and it was not clear how to
extend server configuration with new fields. After the refactoring,
it's clear that adding a new field requires thinking about whether it
should be part of the preconfig or not -- and it could be optional in
the preconfig and mandatory in the config, etc.

### DIFF
--- a/src/main/learnocaml_main.ml
+++ b/src/main/learnocaml_main.ml
@@ -269,17 +269,13 @@ let main o =
        Random.self_init ();
        Lwt.catch
          (fun () ->
-           Learnocaml_store.get_from_file ServerData.enc_init server_config
-           >|= fun pre_config ->
-           match pre_config.ServerData.secret with
-             | None -> None
-             | Some x -> Some (Sha.sha512 x))
+           Learnocaml_store.get_from_file ServerData.preconfig_enc server_config)
          (function 
-           | Unix.Unix_error (Unix.ENOENT, _, _) -> Lwt.return None
+           | Unix.Unix_error (Unix.ENOENT, _, _) -> Lwt.return ServerData.empty_preconfig
            | exn -> Lwt.fail exn) 
-       >>= fun secret ->
-         let json_config = ServerData.default ?secret () in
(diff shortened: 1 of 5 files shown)


yurug:
I am not sure to understand what you mean by `preconfiguration`.

To me, there is a standard notion of `configuration`: a set of parameters whose values modify the behavior of a system globally. A configuration can be modifiable at runtime or not. In our case, the configuration is set once and for all before the server is launched, and never changed at runtime, isn't it?

gasche:
Since #286, there are in fact two formats of learn-ocaml server configuration: those that are expected to be found in `$(repo_dir)/server_config.json`, and those that are exepected to be found in `$(app_dir)/server_config.json`. The former is written by people curating the instance's corpus, and the latter is generated at `learnocaml build`-time by reusing some data from the former, potentially adding some data passed as command-line parameters or generated at build-time. I call the former "preconfiguration" and the latter "configuration". (The name `pre_config` comes from the existing codebase.)

Today (before this PR and also after this PR) the preconfiguration contains the instance secret (used to salt tokens, if I understand correctly), and the configuration also contains the "instance timestamp" (called server_id) that lets us notice when a grader is stale and refresh the information in this case.

The existing code (from #286) claims that these two formats of configuration have the same type, and it uses two different Json serialization formats to deal with the difference between the two types. That approach (is hard to understand and) would not scale to situations where, for example, a type of data could be optional in the preconfiguration and mandatory in the configuration (so we would like to have an `option` type only in the preconfiguration).

gasche:
Ah: I'm doing this work because I want to add a `base_url` parameter to learn-ocaml to be able to run it under a frontend proxy server -- to run several instances under the same virtual machine.

gasche:
(I just rebased on top of the new master; thanks for merging #324!)

yurug:
Thank you @gasche !

PR #327, opened 2019-10-06
[minor] refactor Cmdliner usage

gasche:
The goal is to make it as simple as possible to add new fields to
configuration records (Builder.t, Server.t, etc).

The command-line-interface parts of learnocaml_main entangle the
construction of proper configuration records with unpleasant
global-mutable-variable settings. This PR splits each affected 'term'
value into smaller parts, with one purely in charge of building the
configuration record, and another dealing with the global mutation.

### DIFF
--- a/src/main/learnocaml_main.ml
+++ b/src/main/learnocaml_main.ml
@@ -102,13 +102,18 @@ module Args = struct
       output_json: string option;
     }
 
-    let term =
-      let apply
-          exercises
-          output_json grade_student display_outcomes quiet
-          display_std_outputs dump_outputs dump_reports timeout
-          verbose dump_dot =
+    let grader_conf =
+      let apply exercises output_json =
         let exercises = List.flatten exercises in
+        { exercises; output_json }
+      in
+      Term.(const apply $exercises $output_json)
+
+    let grader_cli =
+      let apply
+          grade_student display_outcomes quiet display_std_outputs
+          dump_outputs dump_reports timeout verbose dump_dot
+        =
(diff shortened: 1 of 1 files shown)


src/main/learnocaml_main.ml
```diff
@@ -120,12 +125,14 @@ module Args = struct
         Grader_cli.dump_dot := dump_dot;
         Learnocaml_process_exercise_repository.dump_outputs := dump_outputs;
         Learnocaml_process_exercise_repository.dump_reports := dump_reports;
-        { exercises; output_json }
+        ()
```
yurug: Why did you not remove this statement?

src/main/learnocaml_main.ml
```diff
@@ -194,11 +207,13 @@ module Args = struct
         Learnocaml_process_playground_repository.playground_dir :=
           repo_dir/"playground";
         Learnocaml_process_exercise_repository.n_processes := jobs;
-        { contents_dir; try_ocaml; lessons; exercises; playground; toplevel }
+        ()
```
yurug: Why did you not remove this statement?

src/main/learnocaml_main.ml
gasche: I guess it's a style thing, but in this particular codebase it felt natural to insist on the fact that running this term returns the `()` value (for example it explains why we match on this value in the `apply` people below). Would you like me to remove it?

yurug:
Thank you @gasche!

PR #328, opened 2019-10-10
[fix|config-issue] Update learn-ocaml.opam.locked

erikmd:
* Remove `conf-libev` (which is only a **depopt** of lwt and would require `apt-get install libev-dev`, which seems unneeded for learn-ocaml).

* Run
    ```
    rm -fr _opam;
    opam switch create . --deps-only --locked;
    opam lock learn-ocaml.opam
    ```
    as some dependencies implied from learn-ocaml.opam were missing...

Cc @yurug


erikmd:
Hi @yurug, best wishes for the new year! :sparkles: :camel: 

Could you take a look at this PR #328 in the upcoming days? this should be an easy one.

yurug:
It is not clear that this a good idea because this dependency to `libev` was introduced in reaction to a scalability problem.

erikmd:
Hi @yurug, OK I see! So, I'll close this PR.
Thanks!

PR #338, opened 2020-01-06
Allow synchronization with no change.

### LINKED ISSUE #337: Fail to view student contributions
Hi,

I fail to view the student contributions when I am in the teacher mode. I can see the registered students but I cannot see the exercises they realized. I guess the students have to make a synchronization and the teacher as well right? But even with the synchronization, it does not work. May be, it is due to some problems of synchronization I have when the Git copy is clean (nothing to commit) and for which I obtain the following message in Learn OCaml : "HTTP error (500): Error: Internal Server Error". Could you help me?

Thanks a lot for your help.

All the best,

David.

yurug:
Should fix #337.

### DIFF
--- a/src/state/learnocaml_store.ml
+++ b/src/state/learnocaml_store.ml
@@ -71,7 +71,7 @@ let with_git_register =
        git ["config";"--local";"user.email";"none@learn-ocaml.org"]) >>=
     f >>= fun files ->
     git ("add"::"--"::files) () >>=
-    git ["commit";"-m";"Update"] >>=
+    git ["commit";"--allow-empty";"-m";"Update"] >>=
     git ["update-server-info"]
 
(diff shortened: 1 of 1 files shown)


erikmd:
Dear @yurug, I've noticed that when students click on `Sync`, as there is no popup or so saying "Sync succeeded", some want to click once again "just to be on the safe side", so that with the latest version, there will be much more (empty) commits.

Do you think this is an issue?

If yes, one can just as well keep the change of #338 as is, and implement a notification popup in another PR? (e.g. with this [MIT-licensed plugin](https://github.com/mouse0270/bootstrap-notify): http://bootstrap-notify.remabledesigns.com, or maybe a [simpler implementation with a manual timeout](https://stackoverflow.com/questions/28796650/how-does-this-popup-appear-and-disappear-after-a-while))
Should I create a separate issue for this idea?

PR #342, opened 2020-01-21
Create an learn-ocaml-client.opam file.

yurug:

PR #344, opened 2020-02-08
Fix CI (missing COPY in Dockerfile)

erikmd:
Hi @yurug, I had only tested #342 locally (without docker) but I guess the Dockerfile had to be updated as well (so this PR should fix the CI build of master that is currently pending).


yurug:
Thank you!

PR #350, opened 2020-05-27
Add bigarray from standard library 

maiste:
Hi,

This PR fixes another part of the issue #241.

It makes the module `Bigarray` available (needed for future work). I have followed what Alexandre did in #260 by keeping the alphabetical order.


yurug:
Could you also add `Float` please?

maiste:
According to the [manual](https://caml.inria.fr/pub/docs/manual-ocaml/libref/Float.html) `Float` is a `4.07.0` module... I don't think I can include it if we want learn-ocaml to stay compatible with Pervasives and OCaml `4.05.0`.

yurug:
OK, we will do that later.

PR #351, opened 2020-05-29
Correct install_printer

maiste:
Hi,
this PR allows the toplevel to support the `#install_printer` directive.

Correct the missing dependency : `Topdirs.printer_type_new`.


yurug:
Thank you!

Issue #305, opened 2019-08-11
First build of learn-ocaml-corpus fails


erikmd:
Following @yurug's suggestion https://github.com/ocaml-sf/learn-ocaml/pull/302#issuecomment-520208261, I've tried compiling learn-ocaml/master `=` 718d1f12ff2ef78b1db34d2f70e4b24ddad0ed53 + learn-ocaml-corpus and noticed a subtle bug: the first build of learn-ocaml-corpus fails, but the second succeeds.

For the record, I summarize below the opam dependencies I used to reproduce this.

<details><summary>OPAM dependencies</summary>

```
$ opam list
# Packages matching: installed
# Name                  # Installed  # Synopsis
asak                    0.1          Partitioning OCaml codes
astring                 0.8.3        Alternative String module for OCaml
base                    v0.9.4       Full standard library replacement for OCaml
base-bigarray           base
base-bytes              base         Bytes library distributed with the OCaml compiler
base-num                base         Num library distributed with the OCaml compiler
base-threads            base
base-unix               base
base64                  2.3.0        Base64 encoding for OCaml
bigarray-compat         1.0.0        Compatibility library to use Stdlib.Bigarray when possible
biniou                  1.2.0        Binary data format designed for speed, safety, ease of use and backw
bos                     0.2.0        Basic OS interaction for OCaml
checkseum               0.0.3        Adler-32, CRC32 and CRC32-C implementation in C and OCaml
cmdliner                1.0.3        Declarative definition of command line interfaces for OCaml
cohttp                  1.1.1
cohttp-lwt              1.1.1        An OCaml library for HTTP clients and servers
cohttp-lwt-unix         1.1.1        An OCaml library for HTTP clients and servers
conduit                 1.3.0        An OCaml network connection establishment library
conduit-lwt             1.3.0        An OCaml network connection establishment library
conduit-lwt-unix        1.3.0        An OCaml network connection establishment library
conf-git                1.0          Virtual package relying on git
conf-libev              4-11         High-performance event loop/event model with lots of features
conf-m4                 1            Virtual package relying on m4
conf-openssl            1            Virtual package relying on an OpenSSL system installation
conf-pkg-config         1.1          Virtual package relying on pkg-config installation
conf-which              1            Virtual package relying on which
cppo                    1.6.6        Code preprocessor like cpp for OCaml
cstruct                 3.3.0        Access C-like structures directly from OCaml
decompress              0.8.1        Implementation of Zlib in OCaml
[...]

PR #353, opened 2020-06-04
Add a regression test for `learn-ocaml build'

agrn:
Commit 5a661e4 ("repo: fix repo generation when an exercise has multiple subdirectories", 2020-05-26) fixed the issue #305, but it did not add any regression test to the test suite to ensure that the behaviour of `learn-ocaml build' stayed correct.

This series cleans up a bit `runtests.sh`, adds such a test, and makes Travis to check `learn-ocaml build' against learn-ocaml-corpus.

This patch series keeps the general behaviour of `runtests.sh` -- perhaps in the future, it could be changed to have a [TAP](https://testanything.org/)-compliant output, or allow it to run without docker?

### DIFF
--- a/tests/runtests.sh
+++ b/tests/runtests.sh
@@ -73,7 +73,7 @@ handle_file () {
 	   learn-ocaml-client grade --json --id="$subdir" \
 	   /home/learn-ocaml/actual/"$subdir"/"$tosend" > res.json 2> stderr.txt
     if [ $? -ne 0 ]; then
-	red "NOT OK: $dir$tosend"
+	red "NOT OK: $dir/$tosend"
 	cat stderr.txt
 	clean_fail
     fi
@@ -88,12 +88,12 @@ handle_file () {
 	    diff res.json "$tosend.json"
 	    # If diff failed
 	    if [ $? -ne 0 ]; then
-		red "DIFF FAILED: $dir$tosend"
+		red "DIFF FAILED: $dir/$tosend"
 		clean_fail
 	    fi
 	fi
     fi
-    green "OK: $dir$tosend"
+    green "OK: $dir/$tosend"
     rm res.json stderr.txt
     let count++
 }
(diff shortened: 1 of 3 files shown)


agrn:
Thank you for your feedback @erikmd!

> 
> LGTM, one remark though: there are several occurrences of the pattern
> 
> ```
> for CORPUS in $(find ./corpuses -mindepth 1 -maxdepth 1 -type d) ; do …; done
> ```
> but this won't work if ever there is a folder that contains a space (I know this is a priori not the case, but better safe than sorry)
> 
> so as suggested in https://mywiki.wooledge.org/BashFAQ/024#Workarounds you may want to phrase your loops like this:
> 
> ```
> while IFS= read -r CORPUS; do …; done < <(find ./corpuses -mindepth 1 -maxdepth 1 -type d)
> ```
> 

Eww, as if I needed to be reminded why I hate shell scripts with a
burning passion.  :)

I am fixing this as I am writing these lines.

tests/runtests.sh
```diff
@@ -42,86 +42,112 @@ learn-ocaml --sync-dir=/sync --repo=/repository build serve")
...
-for DIR in `ls -d */`
+while IFS= read -r dir;
 do
     run_server
 
-    pushd $DIR > /dev/null
+    pushd "$dir" > /dev/null
 
-    echo "---> Doing $DIR:"
+    echo "---> Doing $dir:"
```
yurug, line 124:
```suggestion
    echo "---> Entering $dir:"
```

yurug:
This looks great to me.

Did you try to run [shellcheck](https://github.com/koalaman/shellcheck) on your script?

agrn:
Yes, it reports the usual errors with `pushd`/`popd`, `let`, and backslashes in strings.  Other than that, there is also the only for loop that I did not convert -- for some reason, when converted to a while loop, it will execute only for the first file found.  I am not sure why it behaves like this.

yurug:
Thanks!

Issue #288, opened 2019-06-28
Static deployment of learn-ocaml


erikmd:
Following our latest discussion with @yurug and @leunam217, it seems learn-ocaml is no more deployable statically, i.e., without the ocaml server runtime.

This static-deployment feature could be interesting to allow people put their compiled exercises in a private GitHub repo, and automatically expose the exercises using GitHub Pages (which could just as well be graded at client slide, except that the `Sync` feature would not be available).

Even if this is not high-priority, it may be worth it to have this feature.

erikmd:
FIxed by #356

PR #356, opened 2020-06-25
Fix static deployment

agrn:
@erikmd reported in #288 that it was no longer possible to deploy learn-ocaml statically.  This series fix this by checking the version of the server in the front-end.  If the application is deployed statically, there is no `/version` endpoint, so the front-end will know if it needs to ask the user for a token or not.

The API is also modified so some endpoints can take an optional token, but the server itself will forbid any requests without a token.  There are also some modifications in the repository generation process.

This does not yet address `learn-ocaml-client`, though -- it will fail when trying to talk to a static instance of learn-ocaml.


agrn:
The fail is the macos CI that failed to download some opam packages…

agrn:
I added another patch to fix `learn-ocaml-client`.

erikmd:
Hi @yurug, I've quickly proofread @agrn's PR #356 and it looks very fine.

However, I'd suggest you'd merge PR #353 beforehand because it is also ready and it will extend the CI.

FYI, I've just deployed the result of this PR in a GitHub user Page here:
https://github.com/pfitaxel/pfitaxel-demo-dev/tree/master/docs
and the result is available here:
https://pfitaxel.github.io/pfitaxel-demo-dev

Note that this feature of static deployment was broken at least from https://github.com/ocaml-sf/learn-ocaml/commit/e0a872e5548a3f8116faf664ddac2df36120c896, notably because of the leading "/" in the HTML/CSS/JS URLs, which was de-facto incompatible with having a `BASE_URL` such as `https://pfitaxel.github.io/pfitaxel-demo-dev`.

With this PR, one can deploy a fully working, statically-served instance of learn-ocaml (except the server-side storage feature) incorporating some exercises from a private GitHub repository associated with a public GitHub Pages deployment.

It basically suffices to run:
```bash
$ learn-ocaml build --root "https://pfitaxel.github.io/pfitaxel-demo-dev" --repo=demo-repository
```
(cf. [this example deployment script](https://github.com/pfitaxel/pfitaxel-demo-dev/blob/7042bbd6e00fb85af3b56fcc28d3c480ac2da5a8/deploy#L37) (that should be rewritten to rely on Docker though))

agrn:
Note that it adds a dependency to `markup-lwt` (we already have `markup`, though, so it’s not a big step to take).

erikmd:
@yurug [CI is green now](https://travis-ci.org/github/ocaml-sf/learn-ocaml/builds/703146059), although it seems the automatic PR status feedback from Travis → GitHub doesn't seem to work very well :-/

yurug:
Thank you!

gasche:
If I understand correctly, it is now possible to deploy a learn-ocaml instance as a static website. That sounds like something that could be useful for my own teaching this semester.

Is this static deployment mode documented in `docs/`? I haven't found any information. Could you update `docs/howto-deploy-a-learn-ocaml-instance.md` to document static deployment?

In the meantime, I'm going to experiment with @erikmd's documentation in the comment above ( https://github.com/ocaml-sf/learn-ocaml/pull/356#issuecomment-651116200 ).

Note: last year I tried to deploy several learn-ocaml instances on the same server, using different base URLs. I sent a couple preliminary PRs, but never managed to finish the work (a lot of URLs in various places had to be fixed, etc.). If I understand correctly, doing this is possible with static deployment; this could be very helpful for me.

erikmd:
Hi @gasche

> Is this static deployment mode documented in `docs/`? I haven't found any information. Could you update `docs/howto-deploy-a-learn-ocaml-instance.md` to document static deployment?
> 
> In the meantime, I'm going to experiment with @erikmd's documentation in the comment above ( [#356 (comment)](https://github.com/ocaml-sf/learn-ocaml/pull/356#issuecomment-651116200) ).

just FYI [the `deploy` script](https://github.com/pfitaxel/pfitaxel-demo-dev/blob/master/deploy) I had mentioned in my earlier comment is now migrated to `docker run`, so no opam switch or so is needed anymore to build the `www` folder for learn-ocaml static deployment, and one just need to (install Docker Engine and) run:

```
$ ./deploy /path/to/learn-ocaml-repository
```
to generate the `www` folder, commit, and (if one validates with Return), push.

Anyway, it's true we'll need to document further that feature before next learn-ocaml release! - I'll try to open a pull request.

PR #357, opened 2020-06-25
Move to dune 1.7

maiste:
Hi,
this PR moves the dune-project from 1.6.3 to 1.7.
For a future feature, dune needs to support `dune lang 1.7` to allow to use `virtual_modules` stanza.

This PR changes the following things:
* dependency requirements:
  * dune must be >= dune `1.7`
  * ipaddr is fixed to `2.8.0` to avoid dependency conflicts
  *  lwt is moved to `4.0.0`
* support js_of_ocaml `3.3` and  tyxml `4.4.0` and update files to work with
* remove unused variable errors that stops the compilation with dune new version

### DIFF
--- a/dune
+++ b/dune
@@ -1,4 +1,4 @@
-(ignored_subdirs (docker _obuild www sync demo-repository))
+(dirs :standard \ docker _obuild www sync demo-repository)
 
 (rule
  (targets VERSION)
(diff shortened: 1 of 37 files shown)


yurug:
Thank you @maiste !

Issue #316, opened 2019-09-23
automatic save issue with concurrent edition of the same exercise


erikmd:
**Kind:** bug
**learn-ocaml version:** 0.11

## Context

The latest release of learn-ocaml uses some automatic save of the exercise as soon as the window is closed.

On the one hand, it is not very "intuitive" because the operation is precisely performed transparently for the user (so several students told me this afternoon that they were not sure whether and when their exercise was saved…)

On the other hand, I've noticed the issue below (which we had also noticed and fixed in learn-ocaml-editor (so I guess I could propose a similar PR to adddress this when I'll have some time))

## Steps to reproduce the bug

1. we open in two different tabs or windows of the browser, two views of learn-ocaml with the same user and the same exercise, say, Tab A and Tab B [this can be be useful e.g. if one has a dual screen, namely a laptop screen and an additional monitor].
2. we slightly modify Tab B (without clicking on "Sync" nor "Grade").
3. we thoroughly edit Tab A (say, with several additional lines) and click on "Sync".
4. we close Tab A, *then* Tab B (without clicking on "Sync").
5. we reopen the same URL of the exercise.

## Actual result

- The version of Tab B shows up.

## Expected result

- when closing Tab B at step 4., a standard dialog is displayed, and asks whether to exit without saving the changes; we choose "no" on purpose.
- after step 5, the version of Tab A shows up.

PR #372, opened 2020-12-22
Offer better protections against solution overwriting

yurug:
## Context

This PR introduces three mechanisms to avoid several synchronization issues when two clients are opened on an exercise solution. 

### Mechanism 1
We disable the automatic and implicit saving of the student answer when the browser tab is closed. Instead, we ask the user to confirm that she wants to leave the page.

### Mechanism 2
When an answer has not been modified for 3 minutes, we check if a more recent solution exists on the server. In that case, we ask the user if she wants to download the most recent version.

### Mechanism 3
To avoid to overload the server with many synchronization requests, we disable the synchronization button when the answer is synchronized, and reactive it only when a modification is made on the answer.

Fixes: #316

### DIFF
--- a/src/ace-lib/ace.ml
+++ b/src/ace-lib/ace.ml
@@ -102,7 +102,8 @@ let create_editor editor_div check_valid_state =
   editor##.customData := (data, None);
   editor##setOption (Js.string "displayIndentGuides") (Js.bool false);
   editor##on (Js.string "change") (fun () ->
-      check_valid_state (set_contents data) (fun () -> focus data);
+      check_valid_state (set_contents data) (fun () -> focus data)
+        (fun () -> set_synchronized_status data true);
       set_synchronized_status data false);
   data
 
--- a/src/ace-lib/ace.ml
+++ b/src/ace-lib/ace.ml
@@ -1,6 +1,6 @@
 (* This file is part of Learn-OCaml.
  *
- * Copyright (C) 2019-2020 OCaml Software Foundation.
+ * Copyright (C) 2019-2022 OCaml Software Foundation.
  * Copyright (C) 2016-2018 OCamlPro.
  *
  * Learn-OCaml is distributed under the terms of the MIT license. See the
--- a/src/ace-lib/ace.ml
+++ b/src/ace-lib/ace.ml
@@ -86,6 +86,8 @@ let set_synchronized_status editor status =
   List.iter (fun obs -> obs status) editor.sync_observers;
   editor.synchronized <- status
 
+let focus { editor } = editor##focus
+
 let create_editor editor_div check_valid_state =
   let editor = edit editor_div in
   Js.Unsafe.set editor "$blockScrolling" (Js.Unsafe.variable "Infinity");
@@ -100,7 +102,7 @@ let create_editor editor_div check_valid_state =
   editor##.customData := (data, None);
   editor##setOption (Js.string "displayIndentGuides") (Js.bool false);
   editor##on (Js.string "change") (fun () ->
-      check_valid_state (set_contents data) ;
+      check_valid_state (set_contents data) (fun () -> focus data);
       set_synchronized_status data false);
   data
 
@@ -192,7 +194,6 @@ let clear_marks editor =
 let record_event_handler editor event handler =
   editor.editor##(on (Js.string event) handler)
 
-let focus { editor } = editor##focus
 let resize { editor } force = editor##(resize (Js.bool force))
 
 let get_keybinding_menu e =
--- a/src/ace-lib/ace.ml
+++ b/src/ace-lib/ace.ml
@@ -1,6 +1,6 @@
 (* This file is part of Learn-OCaml.
  *
- * Copyright (C) 2019 OCaml Software Foundation.
+ * Copyright (C) 2019-2020 OCaml Software Foundation.
  * Copyright (C) 2016-2018 OCamlPro.
  *
  * Learn-OCaml is distributed under the terms of the MIT license. See the
@@ -20,6 +20,8 @@ type 'a editor = {
   editor: ('a editor * 'a option) Ace_types.editor Js.t;
   mutable marks: int list;
   mutable keybinding_menu: bool;
+  mutable synchronized : bool;
+  mutable sync_observers : (bool -> unit) list;
 }
 
 let ace : Ace_types.ace Js.t = Js.Unsafe.variable "ace"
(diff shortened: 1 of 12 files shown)


yurug:
@erikmd Can you take a look at this MR?

erikmd:
Hi @yurug, thanks for implementing this!

> Can you take a look at this MR?

Sure! I'll have a look ASAP

src/app/learnocaml_common.mli
```diff
@@ -118,18 +118,23 @@ val set_state_from_save_file :
...
-    file. The save will be created on the server if it doesn't exist.
+(** 
+    [sync token on_sync] synchronizes the local save state with the server state,
+    and returns the merged save file. The save will be created on the server
+    if it doesn't exist. [on_sync ()] is called when this is done.
 
-    This syncs student {b,content}, but never the reports which are only synched
-    on "Grade" *)
-val sync: Token.t -> Save.t Lwt.t
+    Notice that this function synchonizes student {b,content} but not the
```
yurug, line 126:
```suggestion
    Notice that this function synchronizes student {b,content} but not the 
```

erikmd:
Hi @yurug, sorry for the delay! but I should have more time this week, to be able to tackle this :crossed_fingers: :-)

yurug:
@erikmd Any news related to this PR?

erikmd:
Hi @yurug, not yet, sorry; I've been swamped by several exams… but it's still in my backlog!
BTW did you plan to do a release incorporating these features at a specific date?
would it still be OK if I test/review the PR by this WE?

yurug:
@erikmd Will you  have time to review this PR in a close future?

erikmd (commented):
Hi @yurug, belated thanks for this PR! (which addresses a very important usability issue)

I rebased the PR on latest master, eveything compiles cleanly with OCaml 4.12.

Then I tested the app on learn-ocaml's `demo-repository` with firefox-esr on Linux.

The three mechanisms you proposed nicely fit together, and I have one or two remarks/suggestions for each of them:

> ### Mechanism 1
> 
> We disable the automatic and implicit saving of the student answer when the browser tab is closed. Instead, we ask the user to confirm that she wants to leave the page.

Minor issue: when the tab is already synchronized, I think the confirmation dialog `This page is asking you to confirm that you want to leave - data you have entered may not be saved.` should not appear, while it appears unconditionally(?)

> ### Mechanism 2
> 
> When an answer has not been modified for 3 minutes, we check if a more recent solution exists on the server. In that case, we ask the user if she wants to download the most recent version.

Nice enhancement! I noticed that the dialog:

* only shows up when the current tab is already synchronized (i.e., no unsaved code)
* only proposes two choices (basically, "Keep editing" or "Fetch and overwrite")

I believe we could extend the logic and ask:

1. if the tab is already synchronized:  
   (`Fetch from server | Ignore & keep editing | Do nothing` ?)
   → using a dedicated dialog with 3 choices
2. if the tab contains unsaved code:
   (`Ignore & keep editing | Sync & keep editing | Fetch from server & overwrite | Do nothing` ?)
   → using a dedicated dialog with 4 choices, *and* making sure that if the user clicks on anything but `Do nothing`, the dialog won't appear again at the next keystroke.

Would you agree with these two suggestions?

> ### Mechanism 3
> 
> To avoid to overload the server with many synchronization requests, we disable the synchronization button when the answer is synchronized, and reactive it only when a modification is made on the answer.

Good idea! this actually solves "upstream directly, from the frontend" [that issue](https://github.com/ocaml-sf/learn-ocaml/pull/338#issuecomment-572295106) I had raised some time ago.

But small nitpick: I believe the `Sync` button should be greyed at the beginning (when the exercise has just been opened), as there's no new changes to sync at that point… but this is not the case currently.

src/app/learnocaml_common.ml
```diff
@@ -708,11 +712,48 @@ let mouseover_toggle_signal elt sigvalue setter =
...
+    match Learnocaml_local_storage.(retrieve (exercise_state id)) with
+    | { Answer.mtime; solution; _ } ->
+       if mtime > checking_time then (
+         if Js_utils.confirm
+              [%i "A more recent answer exists on the server. \
+                   Do you want to update the current one?"]
+         then
+           update_content solution;
+       );
+       Lwt.return ()
```
erikmd, line 740: Hi @yurug, just in case this is useful for implementing the front-end change (to turn the OK/Cancel JavaScript dialog into a 2+_choice learn-ocaml dialog), here is a working snippet:
```suggestion
         lwt_alert ~title:"Question"
           ~buttons: [ [%i "Fetch"], (fun () -> Lwt.return (update_content solution));
                       (* [%i "Keep Editing"], (fun () -> TODO ); *)
                       [%i "Abort"], (fun () -> Lwt.return_unit) ]
           [ H.p [H.txt [%i "A more recent answer exists on the server. \
                             Do you want to fetch the new version?"] ] ]
       ) else Lwt.return_unit
```

src/ace-lib/ace.ml
```diff
@@ -77,15 +82,34 @@ let get_contents ?range e =
...
+      keybinding_menu = false;
+      synchronized = true;
+      sync_observers = []
+    }
+  in
   editor##.customData := (data, None);
   editor##setOption (Js.string "displayIndentGuides") (Js.bool false);
+  editor##on (Js.string "change") (fun () ->
+      check_valid_state (set_contents data) ;
+      set_synchronized_status data false);
```
erikmd, line 104: BTW note that if we use the snippet suggested in [my other comment](https://github.com/ocaml-sf/learn-ocaml/pull/372/files#r710583507), then because of `lwt_alert`, we should refocus on the editor afterwards, for example here:

```suggestion
      set_synchronized_status data false;
      editor##focus);
```

yurug:
> Minor issue: when the tab is already synchronized, I think the confirmation dialog This page is asking you to confirm that you want to leave - data you have entered may not be saved. should not appear, while it appears unconditionally(?)

I turn this one in a follow-up issue (#467) : I cannot find a way to have firefox stop asking for confirmation.

src/app/learnocaml_common.ml
yurug: Unfortunately, we cannot change the button labels of the confirmation dialog box (as far as I know).

yurug: Done in d90944f

src/ace-lib/ace.ml
yurug: Done in 6b63b7d

erikmd:
Hi, thanks @yurug !

I tested the latest version and committed several minor enhancements.

Before squash-merging:

* I'll try to have a look at #467 (maybe some snippet from the learn-ocaml-editor could help)
* And I'll retest PR on at least two browsers

erikmd:
The fix b9960f747a4b84746e1b3d5cfe9caac6b12f3c78 has been tested successfully with:
- [x] Mozilla Firefox 91.4.1esr
- [x] Chromium 98.0.4758.102 (on Debian GNU/Linux)
- [ ] Safari

erikmd:
Good news: the latest version of the PR seems very fine with (Firefox, Chromium, Safari).
So I'll push a very last comment-cleanup commit and squash-merge.

PR #385, opened 2021-02-22
fix: runtime error in docker packaging

erikmd:
Hi @yurug,

This PR (that makes some alpine dependency explicit) should fix the issue currently observed with the [CI build of branch master](https://github.com/ocaml-sf/learn-ocaml/actions/runs/574189523).

 feel free to merge it if the CI passes anew.

### Details

* Add missing Alpine package
  href: https://pkgs.alpinelinux.org/package/v3.7/main/x86_64/openssl

* Avoid runtime errors:
  Error loading shared library libssl.so.1.0.0: No such file or directory (needed by /usr/bin/learn-ocaml)
  Error loading shared library libcrypto.so.1.0.0: No such file or directory (needed by /usr/bin/learn-ocaml)
  […]

* Ensure this patch is future-proof (instead of adding `libssl1.0`), see: https://github.com/eclipse/mosquitto/issues/1711#issuecomment-637987176

BTW: later on (after some testing), we may also want to bump the version of the Alpine distribution…

PR #388, opened 2021-04-23
Static builds

AltGr:
Adds a simple script that generates completely static builds of `learn-ocaml`, `learn-ocaml-client` and `learn-ocaml-server`.

The building uses Docker, Alpine and Musl, targetting only `x86_64/linux` at the moment.


AltGr:
It's a stand-alone script at the moment, but the process is pretty simple so that shouldn't be hard to include in the CI.

AltGr:
I am puzzled by the CI failure: the patches add a Makefile target but should cause no changes on normal builds :thinking: 
(apart from the relaxed constraint on the dune version, maybe, but the failing CI run uses the latest version anyway)

erikmd:
Hi @AltGr, thanks!

My understanding is that:
- the only job that fails is "not that serious" (it is only the one that does [docker build -t … -f Dockerfile.test-client'](https://github.com/ocaml-sf/learn-ocaml/blob/73ef4c396b2caf710f7be4026f70cef4f941b176/.github/workflows/build-and-test.yml#L42) which just **intends to verify that [the `learn-ocaml-client.opam` file alone](https://github.com/ocaml-sf/learn-ocaml/blob/7c9132dc628a59dea79ce1035fe375941f6fac2f/Dockerfile.test-client#L10) is OK** in terms of dependencies)
- then note that in the repo there is no `learn-ocaml-client.opam.locked` file, so this confirm this test may be more fragile (*maybe we might want to add one such file at some point*)
- also I had experienced a similar weird failure (at linking in another job so it's not necessarily related, but nevermind) that I fixed by bumping the version of ocaml's alpine image that was very old → see PR https://github.com/ocaml-sf/learn-ocaml/pull/385 where I basically replaced `FROM ocaml/opam2:alpine-3.7` with `FROM ocaml/opam:alpine-3.13-ocaml-4.05`
- and finally it seems the initial branch point for #388 is anterior w.r.t. the merge of #385 (and I believe the GitHub Actions workflow do not trigger an implied "hypothetical merge" for the test, unlike Travis e.g.)

All things put together, I'd suggest you rebase your branch on latest master, and I guess it should be very OK then :)

AltGr:
Thanks for the help! I traced it back to the dune version, funny because my intention was to remove the _lower_ constraint on it (for 4.05 compat without falling into the convoluted and costly double OCaml installation), but the previous constraint was actually an `=`, and it seems we fail with too recent versions as well.

That would need to be investigated, but I guess we'll get into it when we upgrade the supported OCaml version anyway; for now I just put back an upper limit.

AltGr:
On static builds for other targets:
- OSX: that will target only x86_64 for now, since we're at the moment stuck with OCaml 4.05 which isn't compatilbe with the newer M1.
- Windows: I tried to run the scripts I had used for cross-compilation a few years back; unfortunately it won't be that easy to make them work. The [`opam-cross-windows`](https://github.com/ocaml-cross/opam-cross-windows) repository I used does not work out of the box anymore (various compilation errors, for example with [gcc10 and older OCaml](https://github.com/ocaml/opam-repository/pull/16583), but there were further ones down the line) ; and it's only from 4.07 on, but if it's just for the client it shouldn't be a huge issue. Native builds through CI may be the better approach there.

AltGr:
Closing as superseded by #391

PR #389, opened 2021-04-28
Improve UX for learn-ocaml logo

YoanwM:
* fix css so it is more visible when the logo is clickable
* Update the logo itself


erikmd:
LGTM, thanks @YoanwM ! I guess I can merge this.

FTR, here is a screenshot of the outcome (note that the pointer is now OK when hovering the logo).

Screenshot-i_PR_389

PR #390, opened 2021-04-30
Make secret field not visible on first connection 

### LINKED ISSUE #303: Don't show Secret field if no secret is configured
If no secret is configured for a server, ideally the Secret field should not be shown when creating a new token so that students are not confused about what this field is for.

YoanwM:
Make secret field not visible on the first connection when the secret is not set on the server

Close #303

### DIFF
--- a/src/main/learnocaml_main.ml
+++ b/src/main/learnocaml_main.ml
@@ -250,7 +250,7 @@ end
 
 open Args
 
-let process_html_file orig_file dest_file base_url =
+let process_html_file orig_file dest_file base_url no_secret =
   let transform_tag e tag attrs attr =
     let attr_pair = ("", attr) in
     match List.assoc_opt attr_pair attrs with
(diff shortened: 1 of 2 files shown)


src/main/learnocaml_main.ml
```diff
@@ -266,6 +266,15 @@ let process_html_file orig_file dest_file base_url =
...
          | `Start_element ((e, "a"), attrs) -> transform_tag e "a" attrs "href"
+         | `Start_element ((e,"div"),attrs) -> if noSecret
+              then (let make_secret_not_visible attrs =
+                      match attrs with
+                      |[] -> []
+                      |((a,b),c)::tail -> if b="id" && c= "secret-section"
+                        then ((a,b),c)::(("","style"),"display:none")::tail
+                        else attrs in
+                    `Start_element ((e,"div"),make_secret_not_visible attrs) )
+              else `Start_element ((e,"div"),attrs)
```
erikmd, line 277: I tested your PR successfully with a secret (or none) in `demo-repository/server_config.json`: the code behaves well at `learn-ocaml build` time (and the static deployment is still OK :).

In order to make the code more concise and more readable, you could use a standard higher-order function such as `List.mem` (i.e., member).

Also you may be interested in using a [guard](https://ocaml.org/releases/4.07/htmlman/expr.html#sec135). FYI, `match l with S e -> if no_secret then e' else S e | t -> t`
can also be written `match l with S e when no_secret -> e' | t -> t`.

So you could use this more compact version of your code:

```suggestion
         | `Start_element ((e, "div"), attrs)
              when no_secret && List.mem (("", "id"), "secret-section") attrs ->
            `Start_element ((e, "div"), (("", "style"), "display:none") :: attrs)
```

erikmd (changes requested):
Thanks @YoanwM! your PR works well, I just left two comments to make the code even more idiomatic :)

You can either refactor your commit locally (by doing `emacs src/main/learnocaml_main.ml` then `git commit -a --amend` and finally `git push -f`) ← recommended approach; or use the automatic buttons `Commit suggestion` in GitHub (also works, but would have the drawback of creating two more commits, instead of amending the single commit you already pushed).

erikmd (approved):

PR #391, opened 2021-05-06
Add github action rules for static binaries generation (Linux, Macos)

AltGr:
Note: this PR supersedes #388.


AltGr:
Aforementioned binaries can be downloaded, in the examples of this run, here: https://github.com/ocaml-sf/learn-ocaml/actions/runs/817393844

Testers welcome!

(note that this might be turned on only for master and tags, but I wanted to demonstrate the PR first :) )

erikmd:
Hi @AltGr, thanks a lot for these two PRs :+1:

I'll test the generated binaries ASAP − at least the Linux ones − and let you know!

scripts/setup-static-build.sh
```diff
@@ -0,0 +1,62 @@
...
+        exit 2
+esac
+
+static_link >src/main/linking_main.sexp laolao_stubs threads camlrun
+
+static_link >src/main/linking_client.sexp threads camlrun
+
+static_link >src/main/linking_server.sexp laolao_stubs threadsnat
+
+echo "src/main/linking_*.sexp generated; remove them to restore a standard (dynamically linked) build." >&2
```
gasche, line 62: I wish those generated files would start with a header explaining that they are machine-generated and what their purpose is. I'm not sure that it works, but would using `; foo` for comments be valid in Dune s-exprs?

gasche: Note: I think that it may be possible to have a better interface, by using dune support for reading environment variables: set things up so that `STATIC_BINARIES=true make` does what we expect, instead of mutating the build environment as is done here.

I think that this could be done by having a rule to build `linking.sexp` as here, with a single rule to produce `linking.sexp` (not a fallback rule and a separate script) that calls a script passing whatever context is needed, plus the content of the STATIC_BINARIES variable. If `STATIC_BINARIES` is "false", the returned sexp is empty.

Sketch:

```
(rule  
  (targets linking_client.sexp)
  (action  (with-stdout-to %{targets}
    (run scripts/linking_flags.sh "client" %{env:STATIC_BINARIES=false}))))
```

AltGr: These options are really specific and not meant for normal uses, which is why I didn't think it would be worth providing a more accessible interface: for example, it won't work for any Linux distribution other than Alpine at the moment.
`make static` will actually start a Docker container to run the build, hence not polluting your workspace.

I am interested about the env variables solution though, it does look much nicer. I couldn't find documentation about `%{env:STATIC_BINARIES=false}`, is that a feature in dune < 2 (to which we are stuck at the moment due to the old OCaml version) ?

gasche: Yes, actually it's available in Dune 1.5.0 or so. See:
- https://github.com/ocaml/dune/issues/1506#issuecomment-774483990
- https://github.com/ocaml/dune/pull/4203

AltGr:
Updated to use the env, thanks @gasche for the suggestion

scripts/static-build.sh
```diff
@@ -0,0 +1,19 @@
...
+set -ue
+
+LC_ALL=C
+
+cd $(dirname "$0")/..
+
+## Run build in container
+
+tar c $(git ls-files) | \
+  docker run --rm -i \
```
erikmd, line 11: Did you have a specific reason for not using a standard "bind mount" here? (I mean, a `docker run -v "$PWD:$PWD" -w "$PWD"`)

Anyway, the current version of this `docker-run` command is OK for me, I guess it is fine to provide more isolation and only share the _build/install/default/bin folder we are interested in.

Though, I'd suggest to tweak the first `tar` command to cope with potential files containing whitespace or so (which is not the case today but might occur some day), and maybe add the `set -o pipefail` so that if the `docker run --rm -i …` command fails but the final `tar vx` does not fail, with this option we are sure to "detect" the failure:
```suggestion
set -o pipefail
git ls-files -z | xargs -0 tar c | \
  docker run --rm -i \
```
WDYT?

scripts/static-build.sh
```diff
@@ -0,0 +1,19 @@
...
+    ocamlpro/ocaml:4.05 \
+    sh -uexc \
+      'tar x >&2 &&
+       sudo apk add openssl-libs-static >&2 &&
+       opam switch create . ocaml-system "dune<2" --deps-only >&2 &&
+       opam exec make LINKING_MODE=static >&2 &&
```
erikmd, line 17: Nitpick (if you don't disagree):
```suggestion
       opam exec -- make LINKING_MODE=static >&2 &&
```

src/main/linking_flags.sh
```diff
@@ -0,0 +1,61 @@
...
+shift
+EXTRA_LIBS="$*"
+
+## Static linking configuration ##
+
+# The linked C libraries list may need updating on changes to the dependencies.
+#
+# To get the correct list for manual linking, the simplest way is to set the
+# flags to `-verbose`, while on the normal `autolink` mode, then extract them
+# from the gcc command-line.
```
erikmd, line 27: Thanks for documenting this.

To ease further updates, do you believe it's feasible to partially automate this? (e.g., adding in the PR a special dune rule and/or make target to print the implied gcc command lines, if an ad-hoc environment variable is set to true)

erikmd (commented):
Hi @AltGr, thanks a lot for this work! (and sorry for the delay… I was quite swamped with teaching this term :-/)

FYI, I've tested successfully the last version of the artifact with:

* GNU/Linux Debian 10 amd64, and
* macOS 11.1, thanks to the help of a colleague of mine.

Apart from minor comments I've left in the review (regarding documentation and bash details), I've identified three points that could be improved, but not necessarily within this PR:

1. The GHA artifacts don't seem to be public (I believe I am able to download them only because I am collaborator of the repo) and according to the docs, they are removed after 90 days, so the way to go would be to rely on the GHA https://github.com/softprops/action-gh-release (mentioned by the legacy GHA https://github.com/actions/upload-release-asset) 

2. For the sake of CI, to detect regressions (namely, if a dependency is missing, and this missing dependency couldn't not be detected without running the binary after compilation), I believe it would be nice to add one job that fetches the artifact using `actions/download-artifact@v2` and runs, say, `learn-ocaml-client --version`; WDYT? and do you believe running the `--version` command would suffice to detect missing deps?

3. For MacOS, my colleague got some warning `Impossible d’ouvrir « … » car le développeur ne peut pas être vérifié.`, as the binary is not signed :-|
however it was fine after running `xattr -d com.apple.quarantine learn-ocaml-client` as suggested [here](https://qastack.fr/programming/60362018/macos-catalinav-10-15-3-error-chromedriver-cannot-be-opened-because-the-de), for example. As neither @yurug nor me have a macOS, I guess it'd be difficult to sign the binaries for the time being, so I'd propose that we document that the macOS binaries are not signed but can be enabled this way with `xattr`…

@AltGr, as the points 1. and 2. above are just related to GHA and can be addressed in another PR (and the point 3. has a workaround), I can propose to look after 1. and 2. myself in another PR if you want; in order to merge this one as soon as you have taken a look at my other few line-comments.

erikmd:
> The GHA artifacts don't seem to be public … and according to the docs, they are removed after 90 days.

Actually this is customizable, but necessarily `<= 90 days`:
2021-07-19_21-03-09_Screenshot_Artifacts_Logs

scripts/static-build.sh
AltGr: Bind-mounts are terrible as soon as your UIDs and GIDs don't match between the host of the container, which is almost always the case, because there is no automatic mapping taking place. Docker advises to use built-in "named volumes" to work around that, but then the data remains hidden somewhere below `/lib` and you need specific commands to extract it. For a simple case like here, this hack seems much nicer ;)

Your suggestion is indeed better in terms of robustness, thanks !

src/main/linking_flags.sh
AltGr: Re-running the (standard) build with `OCAMLPARAM=_,verbose=1` should be enough, find the related `gcc` command towards the end and extract the `-lXXX` arguments.

AltGr:
Feel free to finalise or clarify some comments if you think it's needed before merging; for me it's good to go.

erikmd:
@AltGr (Cc @gasche) just FTR, here is a recap of this PR status and of the few commits I've just pushed, following this Thursday's video meeting with Louis.

First, to follow-up my earlier comment https://github.com/ocaml-sf/learn-ocaml/pull/391#pullrequestreview-709717920 :

> 1. The GHA artifacts don't seem to be public (I believe I am able to download them only because I am collaborator of the repo) and according to the docs, they are removed after 90 days

→ indeed, we could/should rely on https://github.com/softprops/action-gh-release
I'll take care of this before the upcoming release.

> 2. For the sake of CI, to detect regressions (namely, if a dependency is missing, and this missing dependency couldn't not be detected without running the binary after compilation), I believe it would be nice to add one job that fetches the artifact using `actions/download-artifact@v2` and runs, say, `learn-ocaml-client --version`

Actually it seems unneeded to use `actions/download-artifact@v2` because running directly `learn-ocaml-client --version` in the toplevel context within the GHA runner ensures we check the "portability" of the static binary: compiled in alpine (musl-based), then run in Ubuntu Linux.

Anyway I've added a few unit tests, running not only `ldd learn-ocaml-client …` but also `file learn-ocaml-client …` and `learn-ocaml-client --version` (likewise for the two other binaries).

Then, I took the opportunity of this PR to automate what Louis had documented in `src/main/linking_flags.sh`: there is a new Makefile target that prints this 🙂
```
$ make detect-libs  # run on Debian GNU/Linux 10 amd64
...
learnocaml_client: -lcamlstr -lcstruct_stubs -lbase_stubs -lssl_threads_stubs -lssl -lcrypto -llwt_unix_stubs -lpthread -lthreads -lunix -lpthread -lbigarray -lunix -lcamlrun -lm -ldl -lpthread
learnocaml_main: -llaolao_stubs -lbase_stubs -lssl_threads_stubs -lssl -lcrypto -llwt_unix_stubs -lpthread -lthreads -lunix -lpthread -lcamlstr -lcstruct_stubs -lbigarray -lunix -lcamlrun -lm -ldl -lpthread
learnocaml_server_main: -lcamlstr -llaolao_stubs -lbase_stubs -lssl_threads_stubs -lssl -lcrypto -llwt_unix_stubs -lpthread -lthreadsnat -lpthread -lcstruct_stubs -lbigarray -lunix -lm -ldl
```
I have only two remaining questions for @AltGr:

1) among the libs mentioned above, this one is not mentioned [in src/main/linking_flags.sh](https://github.com/ocaml-sf/learn-ocaml/blob/a25955c26b9084d2640c4f27663ed22bba287b0b/src/main/linking_flags.sh#L34-L35) nor in `./dune`: **`-ldl`**. Is this OK anyway? or we should add it in `COMMON_LIBS`.

2) on macOS, the `otool -L` command displays [(link to the logs)](https://github.com/ocaml-sf/learn-ocaml/runs/3135808150?check_suite_focus=true#step:5:442):
```
_build/install/default/bin/learn-ocaml:
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1281.100.1)
	/usr/lib/libncurses.5.4.dylib (compatibility version 5.4.0, current version 5.4.0)
_build/install/default/bin/learn-ocaml-client:
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1281.100.1)
	/usr/lib/libncurses.5.4.dylib (compatibility version 5.4.0, current version 5.4.0)
_build/install/default/bin/learn-ocaml-server:
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1281.100.1)
```
While `/usr/lib/libSystem.B.dylib` corresponds to the macOS standard library that looks mandatory to keep as is, do you think that `/usr/lib/libncurses.5.4.dylib` ought to be "bundled", somehow?

Anyway, as said earlier, we had tested these executables on macOS 11.1 and they were running fine… so maybe a good motto is "don't fix what is not broken…" :)

### To sum up (and sorry for the long post):

I think the PR is ready, so if @AltGr confirms there is nothing further to do for the two points above, I could merge it on, say, this Friday.

erikmd:
following-up my question

> on macOS

about this code
https://github.com/ocaml-sf/learn-ocaml/blob/7276fab2936c6fff0ec6766ecd1b3113d383ff69/src/main/linking_flags.sh#L50

To sum up, thanks to these added lines:
https://github.com/ocaml-sf/learn-ocaml/blob/7276fab2936c6fff0ec6766ecd1b3113d383ff69/.ci-macosx.sh#L15-L16

we get on macOS [(link to the logs)](https://github.com/ocaml-sf/learn-ocaml/runs/3139347029#step:3:1783):
```
learnocaml_client: -lcamlstr -lcstruct_stubs -lbase_stubs -lssl_threads_stubs -lssl -lcrypto -llwt_unix_stubs -lpthread -lthreads -lunix -lpthread -lbigarray -lunix -lcamlrun -lcurses -lpthread
learnocaml_main: -llaolao_stubs -lbase_stubs -lssl_threads_stubs -lssl -lcrypto -llwt_unix_stubs -lpthread -lthreads -lunix -lpthread -lcamlstr -lcstruct_stubs -lbigarray -lunix -lcamlrun -lcurses -lpthread
learnocaml_server_main: -lcamlstr -llaolao_stubs -lbase_stubs -lssl_threads_stubs -lssl -lcrypto -llwt_unix_stubs -lpthread -lthreadsnat -lpthread -lcstruct_stubs -lbigarray -lunix
```

so I guess that we could add `curses` on the two lines calling `./linking_flags.sh`?
https://github.com/ocaml-sf/learn-ocaml/blob/7276fab2936c6fff0ec6766ecd1b3113d383ff69/src/main/dune#L57-L64
(conditionaly if `(= %{ocamlc-config:system} macosx)`, maybe)

AltGr:
> among the libs mentioned above, this one is not mentioned in src/main/linking_flags.sh nor in ./dune: -ldl. Is this OK anyway? or we should add it in COMMON_LIBS.

Well `dl` stands for `dynamic loading`, so... we shouldn't need it :)

> Anyway, as said earlier, we had tested these executables on macOS 11.1 and they were running fine… so maybe a good motto is "don't fix what is not broken…" :)

Yes, apparently you are not really supposed to do purely static executables on macOS, and the system is much, much more homogeneous so we shouldn't need to bother either, our execs should run; the point of statically linking the libs is to avoid runtime dependencies upon optional libs/packages.

> so I guess that we could add curses on the two lines calling ./linking_flags.sh?

IIRC curses is part of the base macOS install so it shouldn't be required, but we could add it (maybe there are older versions still around on some systems ??).
That leaves open the question of why the heck learn-ocaml would require ncurses but well...

erikmd:
Hi @AltGr, thanks for your feedback.

>  That leaves open the question of why the heck learn-ocaml would require ncurses but well...

Maybe because of the `cmdliner` opam dependency?

Also just FYI, I've found [this reference](https://www.oreilly.com/library/view/mac-os-x/0596003560/ch05s02.html) saying:
> **-lm or -lpthread do no harm, but are unnecessary**. The -lm option links to the math library, and -lpthread links to the POSIX threads library. Since libSystem provides these functions, you don't need to use these options.
> …
> In Mac OS X 10.1 and earlier versions, the curses screen library (a set of functions for controlling a terminal display) was part of libSystem.dylib. In Mac OS X 10.2 (Jaguar), the ncurses library (/usr/lib/libncurses.5.dylib) took the place of curses. You may still encounter source code releases that look for curses in libSystem.dylib, which will result in linking errors. **You can work around this problem by adding -lcurses to the linker arguments. This is portable to earlier versions of Mac OS X as well,** …

So as a last tweak, I've removed `pthread` when `%{ocaml-config:system} = macosx`, and I've added `curses` as had suggested the output of `make detect-libs`… If the CI passes anew 😅, I'll merge the current version.

PR #392, opened 2021-05-12
Update the markdown-to-html transformation so links open a new tab

LouisAyroles:
feat: update the transformation from markdown to html which allow to open in a new tab when clicking on a link

### DIFF
--- a/demo-repository/exercises/demo2/descr.md
+++ b/demo-repository/exercises/demo2/descr.md
@@ -1,4 +1,4 @@
-
+[ceci est un lien](http://google.com)
 This exercise is just another demo for the exercise environment.
 <a href="" onclick="top.location='/exercises/demo/';">Test</a>
 
--- a/demo-repository/exercises/demo2/descr.md
+++ b/demo-repository/exercises/demo2/descr.md
@@ -1,4 +1,4 @@
-The following example link will open another tab/window: [OCaml](https://ocaml.org/).  
+The following example link will open another tab/window: [OCaml](https://ocaml.org "External link")  
 This exercise is just another demo for the exercise environment.  
 <a href="" onclick="top.location='/exercises/demo/';">Test</a>
 
--- a/demo-repository/exercises/demo2/descr.md
+++ b/demo-repository/exercises/demo2/descr.md
@@ -1,5 +1,5 @@
-[ceci est un lien](http://google.com)
-This exercise is just another demo for the exercise environment.
+The following example link will open another tab/window: [OCaml](https://ocaml.org/).  
+This exercise is just another demo for the exercise environment.  
 <a href="" onclick="top.location='/exercises/demo/';">Test</a>
 
 <details>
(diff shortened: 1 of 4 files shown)


src/repo/learnocaml_exercise.ml
```diff
@@ -317,8 +317,20 @@ module File = struct
...
+    | Url(href,s,title) ->
+       Some ( let title_url =
+                if title <> ""
+                then
+                  Printf.sprintf {| title="%s"|} title
+                else "" in
+              Printf.sprintf
+              {|<a href="%s" target="_blank" rel="noopener noreferrer"%s>%s</a>|}
+              href title_url (Omd_backend.html_of_md s))
+    | _ -> None in
```
erikmd, line 331: here, I will propose to remove the `let open Omd_representation in` since there is only 1 occurrence of the module;
and whitespace-nitpick: could you reindent/make the code more compact? e.g.:
```suggestion
      let override_url = function
        | Omd_representation.Url(href, s, title) ->
           let title_url =
             if title <> "" then Printf.sprintf {| title="%s"|} title else "" in
           let html =
             Printf.sprintf
               {|<a href="%s" target="_blank" rel="noopener noreferrer"%s>%s</a>|}
               href title_url (Omd_backend.html_of_md s) in
           Some html
        | _ -> None in
```

demo-repository/exercises/demo2/descr.md
```diff
@@ -1,4 +1,4 @@
-
+[ceci est un lien](http://google.com)
```
erikmd, line 1: Good idea to add this test here; but:

* Could you propose a more OCaml-oriented link, and in English? e.g.:
```suggestion
The following example link will open another tab/window: [OCaml](https://ocaml.org/).
```

* BTW did you test what we get for internal links? I mean, `[Internal link](#subsection)` should not generate a link with `target="_blank" etc.` because the href starts with a `#`…

erikmd (changes requested):
Thanks @LouisAyroles !

I left a few remarks/questions in your PR.

demo-repository/exercises/demo2/descr.md
LouisAyroles: Hello @erikmd ,
thanks for your comments. Good idea for a more OCaml-oriented link ! :)

So, you were right, i had to update the code quickly to handle internal links. After few tests, I realize that it was OK but Omd seems to translate weirdly sometimes, for example on the 2 screens below you can see that the "?" is remove from the first id of the title and for the second one, Omd replace the apostrophe by "-39-" and also "é" by "-". 
I don't know if it is caused by HTML which don't allow accent and stuff like this in id.
test
test2

erikmd: @LouisAyroles

> Omd seems to translate weirdly sometimes, for example on the 2 screens below you can see that the "?" is remove from the first id of the title and for the second one, Omd replace the apostrophe by "-39-" and also "é" by "-".
I don't know if it is caused by HTML which don't allow accent and stuff like this in id.

Actually I don't see any issue here: it is very common that the Markdown-to-HTML converters generate an `id=` in pure ASCII without any space… so we could say this is not a bug but a feature ;)

erikmd (approved):
thanks @LouisAyroles, LGTM − let's wait for the CI now − I guess we could merge it this evening.

PR #394, opened 2021-05-14
fix(dune): Avoid ANSI char issues in `opam show … > VERSION`

LouisAyroles:
fix: update dune file to avoid problems with ANSI char with opam show command


erikmd (approved):
Thanks @LouisAyroles, this was a long-standing bug I guess…

Will merge after the CI is green.

PR #396, opened 2021-05-18
fix(omd): Update override_url with htmlentities

LouisAyroles:
The goal here was to handle escaped characters in URL.


erikmd (approved):
Thanks @LouisAyroles, I've tested the code with this example markdown link:
```
[Link](https://fr.wikipedia.org/w/index.php?title=Écho&action=info "External \"link")
```
(with old master, current master, and https://github.com/ocaml-sf/learn-ocaml/pull/396) and that's OK. Merging then

PR #410, opened 2021-07-30
Discourage redundant token creation from users

AltGr:
with less obvious new token registration, students should be much ~more~ **less** likely
to create a new token every time (by mistake or laziness)


erikmd:
@AltGr I believe a small rebase is needed for this PR (and maybe, a translation-file regeneration?)

feel free to ask if ever you'd want that I do it! given it may be a bit cumbersome to do… but quite straightforward at the same time ;)

AltGr:
Rebased

yurug:
Thank you!

PR #411, opened 2021-07-30
QoL: reset the toplevel every time before evaluating code from the editor

AltGr:
You might lose what you manually typed directly in the toplevel pane, but that 
shouldn't be a problem, and most importantly it will avoid already defined ids 
to clutter the environment.

These often cause student confusion with cases like "type t instead of t", or
a missing `rec` keyword that compiles because using an older definition.


erikmd:
> in line with the principle of "least surprise": will the UI show The toplevel has been reset. (* loading the prelude...*), as if the student had clicked on Reset?

Just tested the PR and it's indeed the case:

2021-09-01_02-04-05_Screenshot_PR_411_OK

so I guess I can merge the PR. Thank you Louis for opening this.

Issue #414, opened 2021-08-27
`learn-ocaml build -j 2` is broken


erikmd:
### Bug description

It seems there's a race condition in `learn-ocaml build` that prevents us from safely using the `--jobs` option.

### To Reproduce

* Clone learn-ocaml (`git clone https://github.com/ocaml-sf/learn-ocaml.git`)
* Run `cd learn-ocaml`
* Docker-pull learn-ocaml master (`docker pull ocamlsf/learn-ocaml:master`)
* Run `docker run --rm -it -v "$PWD/demo-repository:/repository" ocamlsf/learn-ocaml:master -j 2`

Then, there is high probability that we get:
```
Learnocaml v.0.12 running.
Updating app at ./www
Cannot process exercises: Unix.Unix_error(Unix.EEXIST, "mkdir", "./www/static")
```
if the repository contains 2+ exercises (which is now the case of demo-repository).

### Expected behavior

No error should happen.

AltGr:
I have examined this a bit, and it attempts to use `Lwt_unix.fork()` which just can't be made to work reliably (see e.g. https://github.com/ocsigen/lwt/issues/812 and there are others like it) ; duplicating the Lwt state in the child process is just going to result in a mess, I am even puzzled that there appears to have been an Lwt version where it worked at some point.

So the option should just be disabled at this point. The more reasonable implementation could be:
- define a dedicated CLI command to compile a specific exercise
- run a whole process with that command inside the `Lwt_list.map_p` on each exercise, using e.g. `Lwt_process.exec`. The difficulty is to make sure that all relevant options in the current context are properly passed to the child process.

Or, of course, we could wait for multicore which should make it fairly trivial :rocket:

erikmd:
Hi @AltGr, thanks a lot for investigating, and for these pointers.

> Or, of course, we could wait for multicore which should make it fairly trivial

Yes, that sounds a good trade-off (and replacing the `learn-ocaml build -j` option with a `-j 1` fallback + warning, for now :)

BTW, I've just skimmed the open issues, and it seems this had been spotted some time ago;
(just putting the URL here to link both issues: https://github.com/ocaml-sf/learn-ocaml/issues/158)

PR #416, opened 2021-09-08
Fix `learn-ocaml build -j 2` issue

erikmd:
This PR is intended to Reproduce and Fix ocaml-sf/learn-ocaml#414

### DIFF
--- a/tests/runtests.sh
+++ b/tests/runtests.sh
@@ -234,7 +234,7 @@ do
 
     if ! ( set -x; docker run --entrypoint '' \
            -v "$(realpath "$corpus"):/repository" \
-           learn-ocaml /bin/sh -c "learn-ocaml --repo=/repository build" ); then
+           learn-ocaml /bin/sh -c "learn-ocaml --repo=/repository build -j 2" ); then
         red "Failed to build $corpus"
         exit 1
     fi


erikmd:
Hi all,

I've tried 2 different fixes in this PR (tweak `Lwt_utils.mkdir_p` (in my reverted commit) or just calling `Lwt_utils.mkdir_p` once more).

Both fix the issue mentioned in `issue #414`: Avoid `Cannot process exercises: Unix.Unix_error(Unix.EEXIST, "mkdir", "./www/static")`

**However**, the issue is more pervasive than what I thought at first sight: despite this potential fix,
the parallel processing is still broken… because we get randomly, with the 2-exercises demo repositiory, one of these outcomes:

1. OK
```
Learnocaml v.0.12 running.
Updating app at ./www
demo2                        [OK]
demo                         [OK]
```

2. Partial build and freeze (the build process hangs):
```
Updating app at ./www
demo2                        [OK]
```

3. something yet more strange:
```
Learnocaml v.0.12 running.
Updating app at ./www
Grader error: Lwt.Resolution_loop.Canceled
demo                       [FAILED]
demo2                        [OK]
demo                         [OK]
```

So, I'm going to postpone this for the time being − fixing this parallel build issue would be nice, but will require more investigation in `src/repo/learnocaml_process_exercise_repository.ml`

(FTR, the outputs above mention "0.12" but the tested version was obviously the commits from this PR;
anyway, I think "the ambient version number" should be bumped in `master` anytime soon, and after each release)

PR #429, opened 2021-10-03
feat: Make server version available to front-ends

erikmd:
This "future-proof PR" is needed to help interoperability in a loosely-coupled way,
so the latest learn-ocaml-client binary (& the learn-ocaml.el front-end depending on it) is compat. with learn-ocaml ↓ version "0.12".

There are now three possible commands:

* `learn-ocaml-client --version`
* `learn-ocaml-client server-version`
* `learn-ocaml-client server-version --min` (the min of the previous two)

Note: the back-end API is unchanged by this PR.


erikmd:
Let's close-reopen this one for CI just in case, as it deals with learn-ocaml-client

PR #430, opened 2021-10-03
fix: details (HTML5 markup)

erikmd:


erikmd:
Merging this one as well (after putting the milestone).

PR #431, opened 2021-10-04
fix: Attempt to workaround the static-bin-macos failure

erikmd:
The current CI failure in master is related to the static-bin-macos job, see https://github.com/ocaml/opam-repository/issues/19676
On second thought, this issue is due to both a broken depext and a (hopefully partly) broken post-check of the presence of the brew package.
So, maybe this PR could workaround the failure…😅


erikmd:
CI is now green! 🙂 So, merging.

PR #439, opened 2021-10-07
A few fixes, in particular for student follow-up on new repos

### LINKED ISSUE #314: Progress bar for students is not displayed on a fresh repository
If a fresh empty /sync is used, the progress bars for students is not displayed.

@nobrakal suggests that this is due to the absence of `exercises.json`. This file can be created by changing the state of exercises from open to close, and to open again. Once this operation is done, the progress bars appear.

AltGr:
Closes #314, #337

Also cleanup the head logos (one camel is enough :) )

And increase the timeout before warning when grading, as requested by some teachers (it happened too fast in normal situations and was distracting to students)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -38,6 +38,7 @@ translations/$(LANGS:=.pot):
 	@for f in $(LANGS); do echo >> translations/$$f.po; done
 	@rm -f translations/*.pot
 	@${DUNE} clean ${DUNE_ARGS}
+	-rm -f ${INDEX_ODOC_PATH}
 	@DUMP_POT=1 ${DUNE} build ${DUNE_ARGS} -j 1
 	@for f in $(LANGS); do \
 	  mv translations/$$f.pot translations/$$f.pot.bak; \


src/state/learnocaml_store.ml
```diff
@@ -203,7 +203,12 @@ module Exercise = struct
           List.iter (fun st -> Hashtbl.add tbl st.id st) l;
           tbl)
       @@ function
-      | Unix.Unix_error (Unix.ENOENT, _, _) -> Lwt.return tbl
+      | Unix.Unix_error (Unix.ENOENT, _, _) ->
+          Lazy.force !index >>= fun index ->
+          Exercise.Index.fold_exercises (fun () id _ ->
+              Hashtbl.add tbl id (Exercise.Status.default id))
+            () index;
+          Lwt.return tbl
```
erikmd, line 211: Hi, thanks a lot @AltGr for your PR!

Just to be sure, do you confirm that this code fixing #314|#337 won't have a side-effect of allowing students that would be already connected to a Learn-OCaml instance URL, to effectively connect to a new server at the same URL if the teacher has reset the `/sync` folder? (because in this case, it would be a small flaw regarding authorization - assuming the instance requires a new "secret" phrase for creating tokens)

Anyway, I think this issue does not occur as the two features are independent, but just wanted to double-check with you 😅

erikmd:
BTW, a small request (sorry for the nitpick):

Could you reword (e.g. with rebase -i) your three commit messages so that they follow [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/)?

E.g.
```
fix(UI): Increase timeout during grading
fix(UI): Cleanup duplicate, inconsistent camel logos
fix(UI): Display exercise status even if the status list hasn't been initialised
```

Because since we use [release-please](https://github.com/ocaml-sf/learn-ocaml/pull/434), this is the way to ensure commits are listed in the automatically-generated [CHANGELOG](https://github.com/ocaml-sf/learn-ocaml/blob/master/CHANGELOG.md)
(we should definitely document this in `CONTRIBUTING.md`, did not have to time to do it yet)

src/state/learnocaml_store.ml
AltGr: This only concerns the exercise status index (as opposed to the student index) so I don't believe it could have any consequence on the tokens. Previously, when the exercise status index didn't exist, exercises would be assumed open, but the index itself was assumed empty, which led to the status of students not being displayed correctly on the teacher screen.

PR #492, opened 2022-06-27
Automate the update of `docs/index.md`

### LINKED ISSUE #490: Feature (Technical Story): Automate the update of index.md
**Related project scope(s):** GHA

**Related user(s):**
- @erikmd

## The problem

The `README.md` and `docs/index.md` can typically be out-of-sync.

### Wanted solution

We should automate the update of `docs/index.md`:

* create a bash script (`#!/usr/bin/env bash`) named `scripts/update-index.sh` that uses `sed` etc. to automatically overwrite `docs/index.md` from `README.md`  
* add a GitHub Action workflow named e.g. `.github/workflows/check-update-index.yml` for each pull request or for master:
  * which clones the repository,
  * runs `scripts/update-index.sh`
  * checks if `git diff` detects that `docs/index.md` should have been modified (committed)
  * in which case the GHA job raises an error (or just a warning if it is master) and suggests to "run `scripts/update-index.sh` and commit"…

cesclass:
* **Kind:** feature

* Close #490 

### Description

Implement a technical (CI) feature to maintain the documentation updated. #490

* Add a script to simplify the update process.
* Fix/Update the `README.md` and `docs/index.md`.
* Add a workflow to check that the index is up to date.

### Checklist

* [x] Read the [CONTRIBUTING.md](https://github.com/ocaml-sf/learn-ocaml/blob/master/CONTRIBUTING.md) guide and:
  * [x] Use [Atomic Commits](https://github.com/ocaml-sf/learn-ocaml/blob/master/CONTRIBUTING.md#atomic-commits) so each commit gathers a single logical change
  * [x] Use [Conventional Commits](https://github.com/ocaml-sf/learn-ocaml/blob/master/CONTRIBUTING.md#conventional-commits) regarding commit messages (needed by our release toolchain)

### Note to maintainers

* Read [this wiki page](https://github.com/ocaml-sf/learn-ocaml/wiki/Checklist-for-testing-and-merging-a-PR)
* Make sure the PR has a milestone
* Assign yourself before merging
* We can squash-merge 1-commit PRs (use a header with a [conventional-commit type](https://github.com/ocaml-sf/learn-ocaml/blob/master/CONTRIBUTING.md#conventional-commits-examples), add a footer with `Fix #…` if need be)

### DIFF
--- a/.github/workflows/check-update-index.yml
+++ b/.github/workflows/check-update-index.yml
@@ -1,0 +1,32 @@
+name: check-update-index
+
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+    branches:
+      - '**'
+
+jobs:
+  check-index:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out the repo
+        uses: actions/checkout@v3
+
+      - name: Execute the script
+        run: ./scripts/update-index.sh
+
+      - name: Check the diff
+        run: echo diff_output="$(git diff --name-only "$PWD/docs/index.md")" >> "$GITHUB_ENV"
+
+      - name: Warning
+        # if diff_output not null and master branch
+        if: env.diff_output != '' && github.ref_name == 'master'
+        run: "echo ::warning:: 'docs/index.md' outdated, please run 'scripts/update-index.sh'."
+
+      - name: Error
+        # if diff_output not null and not master branch
+        if: env.diff_output != '' && github.ref_name != 'master'
+        run: "echo ::error:: 'docs/index.md' outdated, please run 'scripts/update-index.sh' and commit.; exit 1"


scripts/update-index.sh
```diff
@@ -0,0 +1,41 @@
...
+while IFS= read -r line
+do
+
+  # Using sed command with "; semicolon" separator to avoid any problems 
+  #  with the "/ slash" contained in URLs.
+  line=$(echo "$line" | sed -r -e 's;'"$url"';.;g')
+
+  # The read function delete the newline, we need to re-inset it.
+  buffout="$buffout$line"$'\n'
+done <<< "$buffin"
```
erikmd, line 35: That while/read/sed loop looks sub-optimal *and* over-engineered 😅
(as we could just as well have replaced it with a mere `sed -e 's;'"$url"';.;g' "$readme"` command)
but nevermind, this works so let's keep it as is.

scripts/update-index.sh
```diff
@@ -0,0 +1,41 @@
+#!/usr/bin/env bash
+
+# This script update 'docs/index.md' with 'README.md' 
+#  changing github/learn-ocaml/docs URLs to relative path.
```
erikmd, line 4: English nitpick + add a quick motivation:

```suggestion
# This script updates 'docs/index.md' from 'README.md',
# and changes github.com/…/learn-ocaml/…/docs URLs to relative paths,
# so that github.io pages work independently of github.com src files.
```

scripts/update-index.sh
```diff
@@ -0,0 +1,41 @@
...
+
+readme="$PWD/README.md"
+index="$PWD/docs/index.md"
```
erikmd, line 7: Beware that `$PWD/README.md` is not necessarily an absolute path.

To use an absolute path w.r.t. the exact path of the `.sh` script itself (so that the script is robust modulo `cd` commands), I can suggest to write this Bash invocation:

```
srcdir=$(cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null && pwd)

readme=$(readlink -f "$srcdir/../README.md")
index=$(readlink -f "$srcdir/../docs/index.md")
```

scripts/update-index.sh
```diff
@@ -0,0 +1,41 @@
...
+buffout=""
+
+url="https://github.com/ocaml-sf/learn-ocaml/blob/master/docs"
+
+while IFS= read -r line
+do
+
+  # Using sed command with "; semicolon" separator to avoid any problems 
+  #  with the "/ slash" contained in URLs.
+  line=$(echo "$line" | sed -r -e 's;'"$url"';.;g')
```
erikmd, line 31: Always avoid the pattern `echo "$line"`
(because if `line` = `-e` or `-n`, it is eaten…)

Instead, write:
* `printf '%s\n' "$line"`
* or `printf '%s' "$line"`

scripts/update-index.sh
```diff
@@ -0,0 +1,41 @@
...
+
+  # The read function delete the newline, we need to re-inset it.
+  buffout="$buffout$line"$'\n'
+done <<< "$buffin"
+
+# Writing index, deleting the last unnecessary new line
+#  inserted during the process above.
+echo "${buffout::-1}" > $index
```
erikmd, line 39: Sub-optimal echo / `::-1`, and missing double-quote:

```suggestion
printf '%s' "$buffout" > "$index"
```

Indeed, `echo` was adding an extraneous `\n`, while `printf '%s' …` prints the text as is (which already contains the necessary `\n`)

erikmd (changes requested):
Thanks @cesclass !

Your PR is good, but I left a number of details to fix/improve.

Feel free to add commits / `rebase -i` and force-push

scripts/update-index.sh
cesclass: I did not know I can use `printf` instead of `echo` but it seems to be a better way to do.

scripts/update-index.sh
cesclass: I see what you mean for the comsuption of `-e` and `-n`, I will change the approach.

erikmd (approved):
Very good, thanks @cesclass

Issue #521, opened 2023-03-21
Feature: Choose nickname when creating teacher tokens


erikmd:
### Related user(s):

@erikmd

### Related project scope(s):

teacher dashboard

### The problem:

Unlike the student token feature, the teacher dashboard does not ask for nicknames at their creation time.

### Wanted solution:

* Open an input dialog before proceeding to the creation of an extra teacher token?

### Considered alternatives:

* Open a multi-line input dialog (in CSV format or so) to take as input a list of `nicknames, (optional) desired teacher tokens` ?

erikmd:
Closed by https://github.com/ocaml-sf/learn-ocaml/pull/549/commits/c341fcaf6d3da48edceaf37131869d9137277464 from https://github.com/ocaml-sf/learn-ocaml/pull/549

Issue #534, opened 2023-04-07
Bug: the Closed/Open state should appear for Assigned exercises


erikmd:
### Related user(s):

@erikmd, @AltGr

### Related project scope(s):

teacher dashboard

### Bug description:

If an exercise has at least one assigned student, the teacher dashboard displays the `Assigned` state only, while it should display `Open Assigned` or `Closed Assigned`, to show what the exercise status is for unassigned students.

### What OS do you use?

GNU/Linux

### What OS version/distribution do you use?

Debian

### What browser(s) do you use with learn-ocaml?

Firefox, Chromium

erikmd:
@AltGr I keep this issue open because it seemsit is not fixed in master
--- tested with the GHA artifacts https://github.com/ocaml-sf/learn-ocaml/actions/runs/5686137220#artifacts from https://github.com/ocaml-sf/learn-ocaml/pull/557/commits/a1dae0400032aa875093b4e06d3a62e2fd9a9197

Issue #523, opened 2023-03-21
meta-issue: Improve user and developer documentation


erikmd:
### Related user(s):

@yurug, @erikmd

### Related issue(s) or PR(s):

- #504
- #514
- #494 

### Related project scope(s):

documentation

### The problem:

* [ ] @AltGr Reply https://github.com/ocaml-sf/learn-ocaml/issues/588
* [ ] @yurug Document operational semantics of constr of `Learnocaml_api.request`
* [ ] @erikmd Better document availability of static binaries (learn-ocaml serve pas besoin switch) https://github.com/ocaml-sf/learn-ocaml/pull/504
* [ ] @erikmd Document the procedure to fetch/use macOS binaries
* [ ] @erikmd Better document markdown-to-html (feature already in master)
* [ ] @yurug Add Markdown badges in the README?
* [ ] @yurug Mention https://github.com/learn-ocaml-community/teaching-feedback in the README
* [ ] @erikmd Improve dev. documentation (e.g., mentioning direnv, magit, git prompt…)
* [ ] Address https://github.com/ocaml-sf/learn-ocaml/issues/591
* [ ] Address https://github.com/ocaml-sf/learn-ocaml/issues/524
* [ ] Address https://github.com/ocaml-sf/learn-ocaml/issues/573
* [x] @AltGr Add documentation on the teacher board  
  cf. e.g. https://github.com/ocaml-sf/learn-ocaml/issues/514#issuecomment-1412780891

PR #549, opened 2023-05-04
Various improvements to the teacher page, incl. some inline documentation

AltGr:


AltGr:
Hm I got some gettext issues on that last patch, otherwise should be good for review

AltGr:
Closes: #521, #534
Addresses part of #523

src/state/learnocaml_api.ml
```diff
@@ -101,7 +101,7 @@ type _ request =
   | Create_token:
       string * student token option * string option -> student token request
   | Create_teacher_token:
-      teacher token -> teacher token request
+      teacher token * string option -> teacher token request
```
erikmd, line 104: Dear @AltGr, thanks a lot for your very nice PR!

I didn't have the time to compile the branch and test yet (and I have an exam this afternoon at the same time of our usual slot :-/)

but FYI I've only one concern here: at first sight this change is not backward-compatible, cf.

https://github.com/ocaml-sf/learn-ocaml/blob/e060517daecc1f3c5a2fc58223f1d318bd40dd54/src/state/learnocaml_api.mli#L54-L79

Admittedly, the API `Create_teacher_token` is only called from the teacher_tab web frontend, not learn-ocaml-client, but if we we add a create-teacher-token to learn-ocaml some day, it will break compatibility.

Maybe we should discuss this

erikmd: Other idea:
don't change the API constructor `Create_teacher_token`
and do another call to the API to set the nickname, e.g., `Update_save` with a value of type:
https://github.com/ocaml-sf/learn-ocaml/blob/1584a08636eb4ddb875a2b7051f78dbc94bec9b3/src/state/learnocaml_data.ml#L70-L73

src/app/learnocaml_teacher_tab.ml
```diff
@@ -211,7 +211,7 @@ let rec teacher_tab token _select _params () =
             in
             let open_partition_ () =
               Lwt.async (fun () ->
-                ask_string ~title:"Choose a function name"
+                ask_string ~title:"Partitioning of student solutions"
```
yurug, line 255: I am not sure that's a valid patch: we really want to ask a function name here.

src/state/learnocaml_api.ml
AltGr: The type changed at the code level, but the added `string option` is optional in the HTTP request: that means that the newer server version will still handle the requests from an older client in the same way as before.

At least that was the intent, I admit I haven't double-checked ; this probably wasn't included in the CI for the client.

src/app/learnocaml_teacher_tab.ml
AltGr: The question is below, this is only the title of the dialog. My reason for this change is that you get there by ctrl-click on an exercise (i.e. not an explicit button) and previously you may get pretty confused as to what this was about.
image

src/state/learnocaml_api.ml
erikmd: OK! my concern is addressed indeed, LGTM.

erikmd (approved):
Hi @AltGr I tested all the features you implemented in this PR, looks really great!

There's only one issue I spotted (I use firefox-esr on Debian 11):

* https://github.com/ocaml-sf/learn-ocaml/pull/549/commits/03669eb2980fbb42575707e91e50c5be96e4c2dc → the `validate on "enter" keypress` doesn't work for me in the `"Partitioning of student solutions"` dialog, the `"enter"` key is a no-op;
* while for https://github.com/ocaml-sf/learn-ocaml/pull/549/commits/c341fcaf6d3da48edceaf37131869d9137277464 → the `"enter"` key works here.

Can you reproduce the issue?

Apart from that, the PR can be merged as is I think (using a regular merge, **NOT** using a squash).

AltGr:
Ah @erikmd it's weird since it's the same code being called… I re-tested and my guess is that it's probably caused by the popup blocker (I know the browsers limit certain triggers to specific events, like you cannot turn the page fullscreen without user interaction, so maybe a click is required in this case). Here is what I get upon pressing the Enter key:
image

erikmd:
Thanks @AltGr!

Note: don't be surprise that I assign the PR to you after-the-fact; I just follow the convention that PR assignee = merger.

PR #481, opened 2022-04-01
Implement pre-compilation of exercises and graders

AltGr:
Close #478, Close #414, Close #205, Close #312, Close #158

PR #567, opened 2023-10-27
feat(server): add a `--replace` option

### LINKED ISSUE #529: Feature: Ensure `learn-ocaml build serve` immediately handles requests at second run
### Related user(s):

@erikmd

### Related issue(s) or PR(s):

_No response_

### Related project scope(s):

web-app UI, web API, build

### The problem:

* Currently, if we want to change an exercise description or solution or test, we restart `learn-ocaml build serve --repo=…`
* This rebuilds every exercise, which can be time-consuming;
* And in the meantime, the learn-ocaml backend is down (HTTP 502 Bad Gateway or so)

### Wanted solution:

* It would be nice to ensure that the server responds immediately at the second build, even if it needs to rebuild all exos.
* Then when the build is over, **atomically** set the newest directory as the one to be served.

### Considered alternatives:

* Does this require several threads? Would this benefit from OCaml 5?

### Additional context:

_No response_

AltGr:
Closes #529 which seemed to be a common complaint among teachers.

* `learn-ocaml serve --replace` will kill an existing server (running on the
  same port) just before starting

* `learn-ocaml build serve` with an existing server on the same port will fail
  fast (before actually doing the build)

* `learn-ocaml build serve --replace` is more clever:
  - it will do the build **in a temporary directory**
  - then, only if everything is ok, kill the older server
  - swap the files and start the new server

This is all done in order to minimise downtime and be convenient for server
updates.

~~Note that this PR sits on top of #481 and should be rebased once it's
merged.~~

### DIFF
--- a/src/main/learnocaml_main.ml
+++ b/src/main/learnocaml_main.ml
@@ -292,9 +292,15 @@ let process_html_file orig_file dest_file base_url no_secret =
   Lwt_io.close ofile >>= fun () ->
   Lwt_io.close wfile
 
+let temp_app_dir o =
+  let open Filename in
+  concat
+    (dirname o.app_dir)
+    ((basename o.app_dir) ^ ".temp")
+
 let main o =
-  Printf.printf "Learnocaml v.%s running.\n" Learnocaml_api.version;
-  let grade () =
+  Printf.printf "Learnocaml v.%s running.\n%!" Learnocaml_api.version;
+  let grade o =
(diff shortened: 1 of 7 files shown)


erikmd:
Thanks @AltGr, this enhancement is a must-have !!

I didn't finish my tests yet, but as I told you by e-mail I expect this to be done on this Monday Nov 20.

PR #568, opened 2023-11-02
fix(grading): avoid failing on sampling arrays with unique elements

AltGr:
Closes https://github.com/ocaml-sf/learn-ocaml-corpus/issues/39

The sampler was attempting to generate N distinct elements by restarting the
provided sampler an arbitrary number of times (100). It would fail if the domain
is to small: now it might still return duplicates in this case, which is better
than failing for our use-cases.

### DIFF
--- a/src/grader/test_lib.ml
+++ b/src/grader/test_lib.ml
@@ -1762,7 +1762,7 @@ module Make
         if dups then sample else
           let prev = Hashtbl.create max_size in
           let rec sample_new steps =
-            if steps = 0 then invalid_arg "sample_array" else
+            if steps = 0 then sample () else
               let s = sample () in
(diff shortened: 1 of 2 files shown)


AltGr:
Tests fail because of an incompatibility with asak 0.4 (was expecting 0.3), there should be an upper bound. This is unrelated to this PR.

erikmd:
Hi @AltGr,

> Tests fail because of an incompatibility with asak 0.4 (was expecting 0.3), there should be an upper bound. This is unrelated to this PR.

Indeed, but the missing upper bound also prevents master to build (and to do any release), so I merged #570 

Feel free to rebase this PR #568 on master now, and you'll be able to integrate it at the time you see fit.

Issue #366, opened 2020-11-12
The optional file "depent.txt" is not working


git-amd:
Hi,

In the context of some learn-ocaml exercises, I need to preload some external code (".ml" and ".cma" files). I tried to use the optional file "depend.txt" as described in the manual (http://ocaml-sf.org/learn-ocaml/tutorials/step-8.html). However, this seems to fail to work. I cannot manage to make the external code available to the student's code.

Here is the description of one failed experiment. I'm trying to gain access to some module "Q" that exports the symbol "q".

**- Definition of module Q**
> cat my-learn-ocaml-repository/exercises/lib/q.ml

   let q = 123 ;;
   let w = 456 ;;

> cat my-learn-ocaml-repository/exercises/lib/q.mli

   val q: int

**- Checking the module Q**
> ocamlc -c q.mli q.ml
> ocaml q.cmo
        OCaml version 4.05.0
# Q.q;;
- : int = 123
# Q.w;;
Error: Unbound value Q.w

**- The file "depend.txt"**

> cat my-learn-ocaml-repository/exercises/demo/depend.txt
../lib/q.mli
../lib/q.ml

**- Rebuilding the repository and launching the learn-ocaml server**

> touch my-learn-ocaml-repository/exercises/demo/test.ml
> learn-ocaml build  --repo my-learn-ocaml-repository
Learnocaml v.0.12 running.
Updating app at ./www
demo                         [OK]
> learn-ocaml serve
Learnocaml server v.0.12 starting on port 8080
Found the following teacher tokens:
  - X-1JJ-OZA-V41-XXX

**- Using the exercise "demo" in the web browser**

BROWSER:
   Q.q;;
Unbound module Q

Thank you,

- Artur Miguel Dias

yurug:
@lsylvestre Can you help us with this one?

lsylvestre:
Hi, 

thank you for your interest in this feature.
As far as I know, it is not yet possible to preload external code and to make it available to the student's code.

The optional file `depend.txt` just allow to make external code available to `test.ml`, in the grading environment.

For instance, by replacing the content of `my-learn-ocaml-repository/exercises/demo/test.ml` with :

```ocaml
Test_lib.set_result @@ 
  [ Report.info ~message:("Q.q = " ^ string_of_int Q.q) ] ;;
```

the result in the web browser is :

[...]

PR #320, opened 2019-10-03
support for test code reuse

lsylvestre:
**about https://github.com/ocaml-sf/learn-ocaml/issues/169  and  https://github.com/ocaml-sf/learn-ocaml/issues/174.**

This PR offers a support for test code reuse.

It allows the programmer to add an optional file called **depend.txt** (in the **exercises** directory) which declares a list of dependencies as follows :

```
../lib/m1.ml   # just a comment
../lib/m1.mli

../lib/m2.ml
```

Thus, these ml and mli files are copied, ciphered and will be loaded in the toplevel during the grading, just before **test.ml**. It is therefore possible to use the modules *M1* and *M2* in **test.ml** and, furthermore, to use *M1* in *M2*.

PR #578, opened 2023-12-12
remove residual code from PR #320, which is no longer useful

lsylvestre:
PR #481 allows external pre-compiled code to be used in the grading environment, effectively replacing 
the existing mechanism for test code reuse (PR #320). 

This existing mechanism is no longer supported (it is now ignored by grader/grading.ml), 
**but residual code continues to run**. In particular, an exception is raised when file *depend.txt* 
is present in the *exercise* folder and specifies invalid file paths.

This new PR removes residual code from PR #320

This closes issue #366


AltGr:
Thanks @lsylvestre !

PR #601, opened 2024-07-25
Update to OCaml 5.1.1, jsoo 5.8.2

AltGr:
This mainly required time debugging and finding which parts of the code where silently failing with the newer version. Once I found the way to make our version of `ppx_metaquot` work with the newer `ppxlib` (very small code change in the end, but it took quite a bit of digging the ppxlib api), most of the effort was actually in updating jsoo rather than OCaml.

We stick at 5.1 for now because `ppx_tools` (of which we vendor a patched bit) could easily work with it, but isn't ported to 5.2 yet. There has been *lots* of improvements in jsoo in the 5.x branch, and I expect things to be faster although I haven't run benches.

I expect this to be part of a 1.1 release and bumped the version number in prevision


AltGr:
@erikmd I have been running some quick tests locally, but this probably deserves more thorough testing.

PR #602, opened 2024-07-25
Update to OCaml 5.1.1, jsoo 5.8.2

AltGr:
(reopen of #601, closed by mistake)

This mainly required time debugging and finding which parts of the code where silently failing with the newer version. Once I found the way to make our version of ppx_metaquot work with the newer ppxlib (very small code change in the end, but it took quite a bit of digging the ppxlib api), most of the effort was actually in updating jsoo rather than OCaml.

We stick at 5.1 for now because ppx_tools (of which we vendor a patched bit) could easily work with it, but isn't ported to 5.2 yet. There has been lots of improvements in jsoo in the 5.x branch, and I expect things to be faster although I haven't run benches.

I expect this to be part of a 1.1 release and bumped the version number in prevision


AltGr:
Build on OSX fails, but actually at the OCaml compilation stage ; I have no idea about the state of the OCaml 5.1 support, maybe it's been discontinued for x86_64 ?
The CI should probably be updated but we need someone with familiarity with these systems to do so.
Note also that with the new support of opam for Windows, it probably wouldn't be too difficult to propose native Windows builds, as well.

These issues are mostly unrelated to the current PR though.

AltGr:
All other checks pass ; some more manual testing is still warranted (I checked the printing, running and grading only on a few ones)

erikmd:
Hi @AltGr, thanks a lot for this meticulous work!

> Build on OSX fails, but actually at the OCaml compilation stage ;
> I have no idea about the state of the OCaml 5.1 support, maybe it's been discontinued for x86_64 ?

Yes I just saw this failure.
I think it is not an OCaml error, but the fact the `macos-latest` GHA image now only provide arm support:
https://github.com/actions/runner-images/issues/9741

Given the `static-bin-macos` CI job manually installs `opam*x86_64-macos`, I see two approaches:

* either we use the `macos-13` image,
* or we replace the script with installing an arm version of opam/ocaml (this shouldn't induce much change I guess… but maybe reduce the compatibility for old macOS users, won't it?)

> The CI should probably be updated but we need someone with familiarity with these systems to do so.

FWIW I am a macOS user, not a macOS expert but I could test the macOS artifacts anyway!

> Note also that with the new support of opam for Windows, it probably wouldn't be too difficult to propose native Windows builds, as well.

Ah OK, great!
anyway in the meantime, users can rely on WSL.

> These issues are mostly unrelated to the current PR though.

Yes. Except the macOS static binaries I'd say, as it will block continuous deployment and release.

What do you think of just putting `macos-13` here?
https://github.com/ocaml-sf/learn-ocaml/blob/f318b958dae945d7f7e7a7b8fe091f51eb22263e/.github/workflows/static-builds.yml#L75

erikmd:
> What do you think of just putting `macos-13` here?

on second thought: given master already failed this way because of the arm image, I'll push a separate PR with this change.

Then you'll be able to rebase on this small PR (and maybe add `feat:`/`fix:` prefixes to more commits in your PR - those that look important to you)

dune
```diff
@@ -7,7 +7,16 @@
...
-  (release (flags -safe-string -w +a-4-42-44-45-48-3-58-32-33)
+  (release (flags -safe-string -w +a-4-42-44-45-48-3-58-32-33-70)
            (ocamlc_flags)
-           (ocamlopt_flags))
+           (ocamlopt_flags)
+           (js_of_ocaml
+            (compilation_mode whole_program)
+            (flags --no-source-map --opt=2 --enable=use-js-string --target-env=browser)))
+
+  (dev     (flags (:standard -safe-string -w -32 -warn-error -a+8))
```
erikmd, line 17: Two questions:

1. adding this `dev` environment, should I change my usual programming inner loop?

   Namely, coding then building learn-ocaml with a bash alias that runs:
   `cd $dir/learn-ocaml && eval $(opam env) && opam install . --deps-only --locked && make && make opaminstall && learn-ocaml build --repo=demo-repository serve`

2. I had the impression that the dev profile had less warnings enabled than in the release profile ? but probably I'm mistaken

erikmd (commented):
To start with, I just reviewed the opam files. I go on

src/app/dune
```diff
@@ -106,25 +111,25 @@
...
+;;             grading_jsoo
+;;             ace
+;;             learnocaml_repository
+;;             learnocaml_app_common
+;;             lwt_react
+;;             js_of_ocaml-ppx
+;;             ocplib_i18n)
+;;  (modules Learnocaml_partition_view)
+;;  (preprocess (pps ppx_ocplib_i18n js_of_ocaml-ppx))
+;; )
```
erikmd, line 132: should/could this stanza be un-commented,
as a follow-up of https://github.com/ocaml-sf/learn-ocaml/pull/602/commits/787440dabbbdfbc31daae6792c85c682ea9a8936 ?

src/app/dune
```diff
@@ -133,7 +138,7 @@
         (learnocaml_exercise_main.bc.js as www/js/learnocaml-exercise.js)
         (learnocaml_student_view.bc.js as www/js/learnocaml-student-view.js)
         (learnocaml_description_main.bc.js as www/js/learnocaml-description.js)
-        (learnocaml_partition_view.bc.js as www/js/learnocaml-partition-view.js)
+        ;; (learnocaml_partition_view.bc.js as www/js/learnocaml-partition-view.js)
```
erikmd, line 141: should/could this be un-commented as well?

src/main/learnocaml_client.ml
```diff
@@ -812,11 +812,11 @@ module Grade = struct
          (Printf.eprintf "Results NOT saved to server (deadline expired)\n";
           Lwt.return 1)
        else
-         match token with
-         | Some token ->
+         match eo.submit, token with
```
erikmd, line 815: Just out of curiosity, could you elaborate a bit (in a PR comment) on the addition of this condition "eo.submit = true" ?

src/ppx-metaquot/genlifter.ml
```diff
@@ -19,8 +19,21 @@ module Main : sig end = struct
 
   (*************************************************************************)
 
+  module Compat = struct
+#if OCAML_VERSION >= (4, 14, 0)
```
erikmd, line 23: just to be sure, these preprocessor clauses are based on cppo (?)

erikmd (commented):
LGTM, thanks a lot @AltGr for the PR!

FYI next week I'm on vacation - I'm not sure at all I'll be able to perform manual tests by then. But surely in August.

erikmd:
Last question: in the end, would you like that we squash-merge your PR ? or keep all the commits in the master's history.

In the latter case, I'd suggest you `rebase -i` with rewording some commits messages (you know, this is important because they will automatically populate the release changelog).

To make this easier, below are some suggestions for the 31 commits of the PR.

* 21f51fdd *déjà OK* `feat(build): Migrate ocplib-i18n to ppxlib`
* c29d4ae0 **Suggestion=>** `deps(opam): Use pin-depends until release of newer ocp-indent-nlfork`
* 7e744ff8 **Suggestion=>** `build(dune): Update stdlib file names for 4.13`
* f69fb6bf *déjà OK* `fix(build): Fix/disable some warnings`
* 82c1ca79 **Suggestion+Typo**=> `fix: Update genlifter.ml from upstream ppx_tools
* 6a4e595f **Suggestion=>** `deps(opam): Remove dependency on ocaml-migrate-parsetree`
* e7651067 **Suggestion=>** `feat: Upgrade to OCaml 4.13`
* c40c8807 **Suggestion+Typo=>** `feat: Update to OCaml 4.14.2`
* 3b5835a9 **Suggestion=>** `refactor: Disable Asak for now (it's not compatible with OCaml 5.0 at the moment)`
* 28139442 **Suggestion=>** `build(dune): Update dune-project`
* 97d97836 **Suggestion=>** `deps(opam): ocplib-json-typed is now json-data-encoding`
* f5fd701b **Suggestion=>** `deps(opam): Update opam files for OCaml 5.0`
* 6b4289e5 **Suggestion=>** `build(dune): ppx_ocplib_i18n doesn't cope well with dune sandboxing`
* 2546d479 **Suggestion=>** `refactor: Disable partitioning functionality for now (asak isn't compatible with 5.0)`
* 3f1d191e **Suggestion=>** `build(duine): Update stdlib cmi list`
* 64b49681 *déjà OK* `Now compiles (but is broken) with 5.0.0`
* b9dc09c2 **Suggestion=>** `fix: a bunch of deprecation warnings & errors`
* 6032da4c **Suggestion=>** `fix(dune): Adjust jsoo compilation flags`
* 5d555bd3 **Suggestion=>** `deps(opam): Working with OCaml 5.0.0 / jsoo 5.0.1`
* 0217355f **Suggestion=>** `deps(opam): Update jsoo to 5.1.1`
* 378c951f **Suggestion=>** `fix: Many fixes and update for jsoo 5`
* 3641979e **Suggestion=>** `fix: Actually bind the `--dry-run` flag`
* 993e286e **Suggestion=>** `refactor: Improve web-worker debug messages`
* b6c9cc0c **Suggestion=>** `feat: Adjust to OCaml 5.1.1`
* ebd00919 *déjà OK* `feat: complete porting to OCaml 5.1 and jsoo 5.8`
* e87ad942 *déjà OK* `feat: bump version to 1.1.0`
* 787440da *déjà OK* `feat: Reenable asak (using pin-depends for compat patch with OCaml 5)`
* fc6fafe0 *déjà OK* `fix(CI): fix Dockerfile and client dependencies`
* c10038b8 *déjà OK* `fix(CI): use older dependency releases in lockfile`
* 5c5a1332 *déjà OK* `fix(CI): fix static builds and some specific deps`
* 73423d10 *déjà OK* `fix(CI): fix base Alpine version for final Docker images`

learn-ocaml-client.opam
```diff
@@ -29,24 +28,24 @@ depends: [
...
-  "ocaml" {(>= "4.12") & (< "4.13~")}
+  "ocaml" {(>= "5.1") & (< "5.2~")}
   "ocamlfind" {build}
-  "ocaml-migrate-parsetree" {= "1.8.0"}
   "ocp-indent-nlfork"
-  "ocplib-json-typed" {>= "0.7"}
+  "json-data-encoding" {>= "0.7"}
   "ocp-ocamlres" {>= "0.4"}
-  "omd" {<= "1.3.1"}
+  "omd" {<= "2.0"}
```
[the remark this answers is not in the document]
AltGr: > {<= "2.0"} includes omd.2.0.0~alpha4

That's not the case; `<= "2.0.0"` would include it though.
Nevertheless, you are right to point out that this `<=` is meaningless and confusing ;)

learn-ocaml-client.opam
```diff
@@ -18,7 +18,6 @@ homepage: "https://github.com/ocaml-sf/learn-ocaml"
 bug-reports: "https://github.com/ocaml-sf/learn-ocaml/issues"
 dev-repo: "git+https://github.com/ocaml-sf/learn-ocaml"
 depends: [
-  "asak" {>= "0.4"}
```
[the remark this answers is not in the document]
AltGr: ah I think I needed it in the end because `learnocaml_data` contains a type that is defined there, so not re-adding it to the plain `.opam` file was probably a mistake ?
This dependency could well be avoided, but I'm not sure the gain would be worth the added code complexity (needing to split the datatype modules, etc.)

src/app/dune
AltGr: Oops, it definitely should indeed. It's actually weird that I could verify the partition view from the web-app worked without this... did I forget to commit something ? I'll check again.

src/main/learnocaml_client.ml
AltGr: See related commit message (commit messages are better than PR comments IMHO, since they live in the project rather than on a proprietary platform)
This is actually the CLI flag that is the negation of `--dry-run`

src/ppx-metaquot/genlifter.ml
AltGr: Yup!

AltGr:
I'd prefer it not to be squashed as this loses a lot of information. In particular, if one wanted a build working with 4.14, it can easily be extracted now by selecting the correct intermediate commit, and that would be completely lost if squashed.

I don't believe, however, that most of these commit belong to the changelog: saying that we updated jsoo and OCaml versions is the gist of it (plus a few aside improvements, like the `--dry-run` flag fix). For example, it doesn't hold much meaning to have a changelog stating that Asak was disabled and then re-enabled.

dune
AltGr: Historically it was impossible to build in `dev` mode because (i) dune didn't allow to customise the jsoo flags manually, it was tied to the built-in profiles, and (ii) source-map was blowing on us and the compilation wouldn't terminate. Thankfully both of these are now fixed :)

In the new setting, the jsoo `--opt=2` flag makes release compilation quite a bit slower so I'd advise to use the dev mode when, well, developping 🤯 . The `Makefile` still defaults to `release` though, I'll add a `PROFILE ?= release` variable.
Then you could just run `make testrun PROFILE=dev`

2. hm it's quite possible, I haven't taken much attention to the specific warning flags ; I've copied them from another of my projects mainly to disable warn-error in dev mode which I find very hard to work with ("yes I know that this argument is unused, I just commented out the code, will you go on anyways ?" ;) )

learn-ocaml-client.opam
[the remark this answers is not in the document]
AltGr: Ah actually it's there, just got moved around the list

AltGr:
Thanks a lot for reviewing, this last patch should address your concerns :)

src/main/linking_flags.sh
```diff
@@ -68,7 +68,7 @@ case $(uname -s) in
         esac
         ;;
     Darwin)
-        COMMON_LIBS="camlstr ssl_stubs /usr/local/opt/openssl/lib/libssl.a /usr/local/opt/openssl/lib/libcrypto.a cstruct_stubs bigstringaf_stubs lwt_unix_stubs unix"
```
AltGr, line 71: Darwin compilation failed again later on... probably something to adjust here ? I'd be glad @erikmd if you could have a look 🙏🏿

erikmd: Hi @AltGr, sure! I'll be able to take a look (basically, relying on the `make detect-libs` target I had (co)written)
but I'm going into vacation so I won't be able to do it before the week of 12 August… stay tuned, and see you!

AltGr:
This has been long enough, I think this should be merged even with the failure which is limited to the macos scripts;
Unless there are comments within the next days ?

erikmd:
Dear @AltGr,
Best wishes for the new year!

> This has been long enough, I think this should be merged even with the failure which is limited to the macos scripts;
Unless there are comments within the next days ?

Apologies for the ridiculously late review for this PR...
Also, thank you Louis for posting this follow-up comment, without forcibly merging the PR without notice.
So I'm working on the remaining issue tonight.

erikmd:
I try rebuilding the branch on a fresh opam switch with `--deps-only --locked` on a macOS version 14.5. 

The bad news is that the compilation stalls with something like:
```
ld: warning: ignoring duplicate libraries: '-lunixbyt'
Done: 98% (1439/1462, 23 left) (jobs: 8)
```
with no progress despite waiting a dozen of minutes.

I don't know if this issue is the same as the issue you got with the CI.
Anyway, I'll push a temporary debugging commit in your branch to make the macOS build more verbose in the CI.

erikmd:
Hi @AltGr, good news! the branch is ready to merge from my point of view. Thanks again for this important PR!

I just let you take a look at my patches before that if you wish (otherwise, feel free to merge the PR now).

A few comments meanwhile:

1. my tests confirmed that the implementation of a previously-discussed regarding CI and opam dependencies (namely, having jobs that test-build learn-ocaml (1) with the `--locked` dependencies & (2) with the latest compatible dependencies) was a great idea, and made it possible to detect several issues
2. I tried to keep the guideline of **atomic commits**, while in particular the static-build-linking-flags fix is **split in two commits** on-purpose, as I suspect the commit https://github.com/ocaml-sf/learn-ocaml/pull/602/commits/6be6e54e39c156784bfa86409dfb42f10a4db1d5 might be reverted later somehow if `-lzstd` can be dropped when migrating to ocaml 5.2, see also the issue https://github.com/ocaml/ocaml/issues/12562
3. I added https://github.com/ocaml-sf/learn-ocaml/pull/602/commits/de6c0def3e11f06ee0bada49706fdfb988ac7847 with a debug-purpose only at first, but I propose to keep it as is given this verbosity flag is a minimal one, and it may be puzzling to have no output at all with a ~10' build time!
4 quick question for you: with `dune build --display=short`, is the displayed step run after it is displayed, or before? (I believe after, but I'd like to be sure ;)
5. finally regarding performance, I noticed the last js_of_ocaml build steps are a bit long, but given for example that a full build (`% make clean ; time make`) on Apple M1 + macOS 14.5 takes ~8', I think it's fine anyway!
  ```
  % time make
  (...)
  make  952,03s user 682,79s system 344% cpu 7:54,05 total
  ```

See you soon.

erikmd:
@AltGr Last question for you! → what is the status of these two dependencies?

https://github.com/ocaml-sf/learn-ocaml/blob/50d5182db2c0cca9d41e2f11090a0228f418e0e3/learn-ocaml.opam#L86-L95

https://github.com/ocaml-sf/learn-ocaml/blob/50d5182db2c0cca9d41e2f11090a0228f418e0e3/learn-ocaml-client.opam#L59-L64

AltGr:
Rebased and removed the pin-depends ; note that ocp-indent-nlfork is still pending publication (https://github.com/ocaml/opam-repository/pull/27305) so CI is expected to fail until then

erikmd:
Thanks a lot @AltGr for your feedback and for the rebase!

I propose to:

* merge your PR as soon as the CI is green (thanks for handing the point about the dependencies)
* check that the docker image for master builds successfully, then release 1.1.0 myself

Fine with you?

After the release, feel free to make the announcement on Discourse OCaml and `[caml-list]` if you wish (given you're the main contributor for 1.1.0 !) or let me know if you prefer that I look after this - there's an old template [here](https://github.com/ocaml-sf/learn-ocaml/wiki/How-to-announce-new-LearnOCaml-releases)

erikmd:
Regarding the last failing CI job, it seems that the image `ocamlpro/ocaml:5.1` would need a rebuild or so (?)

Other minor questions (follow-up of the upcoming merge of this PR):

* regarding the js_of_ocaml performance and your remark "Maybe we can try fiddling with the options and see if there is no unneeded debug flags left":  
  would you be OK that we open a side issue, just to remember checking this later? (even if I'm not worried at all about this: I believe the current build time is pretty good);

* just to understand better the issue I workarounded in https://github.com/ocaml-sf/learn-ocaml/pull/602/commits/6f6be0f6c80fa27a002c0268d087fe0f37965f2f:  
  do you have an idea why the final js_of_ocaml command includes a **duplicate flag `--no-source-map`** for almost each .js compilation?  cf. `(cd _build/default/src/app && /home/opam/.opam/5.1/bin/js_of_ocaml --no-source-map --no-source-map --opt=2 …` in the dune log; although I don't see the flag duplicated in the dune file…

erikmd:
Hi again, as the [ocamlpro/ocaml](https://hub.docker.com/r/ocamlpro/ocaml/tags) image was rebuilt this afternoon,
I restarted the CI which was successful! So, merging now.
Still, I'll wait for green-light (notably, if you have feedback w.r.t. my previous two comments) before I make the release.

AltGr:
Thanks a lot!

> would you be OK that we open a side issue, just to remember checking this later? (even if I'm not worried at all about this: I believe the current build time is pretty good);

Yes, let's keep track of this

> do you have an idea why the final js_of_ocaml command includes a duplicate flag --no-source-map for almost each .js compilation?

Not sure, but my guess would be that `dune` could have changed its default flags ? Historically jsoo flags couldn't really be customised and I had to force out of the `dev` profile to disable source maps (that would blow up the compiler) ; from [the doc](https://dune.readthedocs.io/en/stable/jsoo.html#sourcemap) sourcemaps are still disabled by default in release mode.

